PR #123, opened 2020-07-03
Fix a few typos, add reference to mypy for config check

clintgibler:
Fixed a few small typos, added a link to installing mypy, as at least on macOS, I received a warning about it when running `promnesia config check`.


karlicoss:
Thanks!

Issue #122, opened 2020-07-02
Issue with running first time index on macOS


flinge:
I've installed libmagic as per [https://filemagic.readthedocs.io/en/latest/guide.html](https://filemagic.readthedocs.io/en/latest/guide.html)

```
denis@MacBook-Pro ~> promnesia index
/Users/denis/Library/Python/3.7/lib/python/site-packages/promnesia/kython/klogging2.py:27: UserWarning: You might want to install 'logzero' for nice colored logs!
  warnings.warn("You might want to install 'logzero' for nice colored logs!")
[INFO    2020-07-02 14:57:20,143 promnesia common.py:417] extracting via promnesia.sources.auto:index ('/usr/share/doc/python3/html/faq',) {} ... ...
[INFO    2020-07-02 14:57:20,143 promnesia common.py:417] extracting via promnesia.sources.auto:index ('/usr/share/doc/python3/html/faq',) {} ... ...
[ERROR   2020-07-02 14:57:20,149 promnesia common.py:433] indexer emitted exception
Traceback (most recent call last):
  File "/Users/denis/Library/Python/3.7/lib/python/site-packages/promnesia/sources/auto.py", line 291, in _index
    yield from _index_file(pp, opts=opts)
  File "/Users/denis/Library/Python/3.7/lib/python/site-packages/promnesia/sources/auto.py", line 317, in _index_file
    pm = mime(pp)
  File "/Users/denis/Library/Python/3.7/lib/python/site-packages/promnesia/sources/auto.py", line 26, in mime
    return _magic().from_file(str(path))
  File "/Users/denis/Library/Python/3.7/lib/python/site-packages/promnesia/sources/auto.py", line 22, in _magic
    return magic.Magic(mime=True)
TypeError: __init__() got an unexpected keyword argument 'mime'

[ERROR   2020-07-02 14:57:20,149 promnesia common.py:433] indexer emitted exception
Traceback (most recent call last):
  File "/Users/denis/Library/Python/3.7/lib/python/site-packages/promnesia/sources/auto.py", line 291, in _index
    yield from _index_file(pp, opts=opts)
  File "/Users/denis/Library/Python/3.7/lib/python/site-packages/promnesia/sources/auto.py", line 317, in _index_file
    pm = mime(pp)
  File "/Users/denis/Library/Python/3.7/lib/python/site-packages/promnesia/sources/auto.py", line 26, in mime
    return _magic().from_file(str(path))
[...]

Issue #20, opened 2019-12-27
real time indexing


karlicoss:
E.g. something inotify based. That would make the implementation quite a bit more complext that it's at the moment.
Also due to the nature of many exports (periodic), it won't be realtime unless the underlying exports are realtime.
Still it could at least detect source files changes, etc.
Also would work well in conjunction with Grasp.

karlicoss:
Might need to be careful about closing libmagic https://github.com/karlicoss/promnesia/pull/124#issuecomment-653890755

karlicoss:
Relevant: i've implemented  'almost realtime' indexing recently: https://github.com/karlicoss/promnesia/blob/c442081c3ac46859462ccd10f65dc89e39a2f44d/src/promnesia/dump.py#L18

E.g. you can have a separate config file only with your text notes (which should be indexed very fast). Then if you run 

`PROMNESIA_INDEX_POLICY=update promnesia index --config /path/to/small/config`,  it will merge it into the main database.

That means you can run it very often (e.g. every five minutes), or potentially combine with `entr` to achieve 'realtime' indexing..

ankostis:
The last comment here needs to make it into main docs.

Even better, if a new option is added like `promnesia index --update` so that the above preserves existing items in server's database:

    promnesia index --update --config <small-config> --secrets <secret-file>

But what about de-duplication?  Are there any issues with updates?

karlicoss:
Yep, good idea to pass it in cmdline args! It was somewhat experimental at first, so I made it an env variable, but it seems to work pretty well (apart from one minor race condition I might need to fix first). 
Maybe even it makes sense to make `--update` mode the default? I guess the worst that would happen is some stale entries would be in the database -- then if the user notices them, they can do a full reindex manually.

Regarding deduplication -- not sure what do you mean? 
This is how it works at the moment
https://github.com/karlicoss/promnesia/blob/e3b21cb080fa9965802bfd2e931ef4263e3a34e9/src/promnesia/dump.py#L61-L79

So it clears all the entries corresponding to the data source first and then inserts them. Hopefully shouldn't result in duplication!

karlicoss:
hmm seems that it was closed automatically by github -- we don't really have realtime indexing yet, so I'll reopen

karlicoss:
Perhaps for actual 'realtime' this would need proper HPI support. 
E.g. HPI module exposes a generator or something, which Promnesia can poll on (presumably, in a loop over all promnesia sources).
[...]

PR #124, opened 2020-07-03
Update filemagic to new API to fix meme crash

clintgibler:
When running `scripts/promnesia index` on macOS, I received the following error:

```bash
TypeError: __init__() got an unexpected keyword argument 'mime'
```

From [src/promnesia/sources/auto.py#L22](https://github.com/karlicoss/promnesia/blob/3da00f34163530e28d0a441ab903e3fae392429c/src/promnesia/sources/auto.py#L22):

```python
@lru_cache(1)
def _magic():
    import magic # type: ignore
    return magic.Magic(mime=True)
```

Reviewing the [filemagic docs](https://filemagic.readthedocs.io/en/latest/guide.html#usage), as linked by @flinge in #122, I was able to fix the error by changing the arguments to the `magic.Magic` constructor and use `magic.id_filename()` instead of `magic.from_file()`.

### Notes

I'm not familiar with the Promnesia code base nor `filemagic`, so honestly I'm not sure if this code behaves as it should. Would greatly appreciate your feedback 🙏 

Also, from the [filemagic docs](https://filemagic.readthedocs.io/en/latest/guide.html#usage),

> To ensure that resources are correctly released by magic.Magic, it’s necessary to either explicitly call close() on instances, or use `with` statement.

Since `_magic()` returns an instance of `magic.Magic` without a `close()` or using `with`, the code may currently be leaking file descriptors, though I'm not sure. 

It would probably be good to use a `with` statement unless there's some huge performance or other impact I'm not aware of. This PR does not use `with` because I was trying to make the diff as small as possible.

### DIFF
--- a/src/promnesia/sources/auto.py
+++ b/src/promnesia/sources/auto.py
@@ -19,11 +19,11 @@ from ..common import Visit, Url, PathIsh, get_logger, Loc, get_tmpdir, extract_u
 @lru_cache(1)
 def _magic():
     import magic # type: ignore
-    return magic.Magic(mime=True)
+    return magic.Magic(flags=magic.MAGIC_MIME_TYPE)
 
 
 def mime(path: PathIsh) -> str:
-    return _magic().from_file(str(path))
+    return _magic().id_filename(str(path))
 
 Ctx = Sequence[str]
 


karlicoss:
Thanks! Appreciate the fix, but let me know if using `python-magic` solves this for you https://github.com/karlicoss/promnesia/issues/122#issuecomment-653890189

But well spotted about closing -- it seems that `python-magic` doesn't implement a context manager, even though in principle it probably should (it relies on `__del__` at the moment https://github.com/ahupp/python-magic/blob/master/magic.py#L118-L131). I guess this is not super critical though -- even if it's leaking some resources, they will be released when indexing is finished along with the whole process. Perhaps this will matter more later if realtime indexing is added.

clintgibler:
It does! I commented as well: https://github.com/karlicoss/promnesia/issues/122#issuecomment-653934805. Closing this PR.

I see you beat me to adding a link to this on https://github.com/karlicoss/promnesia/issues/20 :)

Issue #125, opened 2020-07-04
User workflow documentation / understanding how components fit together


clintgibler:
First off, thanks so much for creating promnesia, HPI, grasp, and more! 🙏 

The past few months, I've been searching for tool(s) that fit my goals for storing and easily accessing information, and reading your [promnesia blog post](https://beepb00p.xyz/promnesia.html) was like reading a better, fleshed out version of what I had in my head  🙌  

I'm starting to play around with the awesome ecosystem of tools you've built, but a few things currently aren't clear.

Apologies for the length of this issue 🤣   I wanted to give you some background of what I'm looking to do and my current understanding of how your various libraries work, so that hopefully the documentation can be fleshed out and made clearer for myself and others. 

Happy to help document based on my experiences using it, if that'd be helpful.

## Context: My Goals

To give you a bit of context, here's what I'm currently doing and what I'm looking to accomplish.

**Info sources** I want to store: 
* Meeting notes, personal notes, snippets from emails
* Conference talk notes
* Highlights and notes from PDFs, web pages, Twitter
* Articles and their highlights from Pocket

**Storage details**:
* I'd like the canonical source of all data to be in org mode files in one git repo.
* The notes will likely be structured using some combination of [org-roam](https://org-roam.readthedocs.io/en/master/) and [PARA](https://fortelabs.co/blog/para/).

**Functionality/workflow goals**:
* I'd like to be able to easily and automatically ingest info from all of the above sources and import them into the central org mode repo.
* When I browse the web, what I've already seen and highlighted should be easily and automatically surfaced, and it should be easy to save/highlight new things.

## How it All Fits Together

I think pretty much all of the above can be done (and you're already doing) with the assorted tools and libraries you've built, per your [personal infrastructure diagram](https://beepb00p.xyz/myinfra.html).

But please permit me to summarize here to make sure I'm understanding things correctly:
* [Human Programming Interface (HPI)](https://github.com/karlicoss/HPI) - Helps export data from a variety of third party services and stores them locally, as JSON/XML/etc.
  * Supports interacting with the downloaded data from any of the supported services as Python objects.
* [orger](https://github.com/karlicoss/orger) - Given input from HPI, enables exporting the data to org mode files.
  * Not totally sure about ^, but guessing based on [this](https://beepb00p.xyz/orger.html#why_hard).
  * Supports parsing org mode files and accessing them as Python objects
[...]

PR #127, opened 2020-07-05
Add hotkey details to FAQ

clintgibler:
Make it easy for users to discover which hotkeys promnesia registers and for what purpose, as mentioned here: https://github.com/karlicoss/promnesia/issues/125#issuecomment-653897258.


karlicoss:
nice, thanks!

PR #136, opened 2020-08-28
forensicswiki link is down, use wayback machine

purarue:
The forensicswiki link describing the desktop firefox sqlite history schema is down, edited the link to go to a recent wayback machine capture


karlicoss:
Thanks!

PR #187, opened 2020-12-06
Documentation Fix: Correct Firefox/Chrome which should be Chrome/Firefox

sandersantema:
This might be some quirk regarding my setup but for me it's Ctrl+Alt+E which for instance opens the sidebar which means the correct order is Chrome/Firefox such that Ctrl+Alt+E corresponds to Firefox instead of Ctrl+Shift+E.


karlicoss:
Oh, indeed, you are correct! https://github.com/karlicoss/promnesia/blob/b9e59262b2fc51ef526248ba273436b0debdedaa/extension/webpack.config.js#L34
Thanks!

Issue #195, opened 2021-02-15
REF: Add tags into context


ankostis:
I see that hypothesis source does not harvest tags, correct?
Although this is fixable, it made me wonder whether tags should be a new column in promnesia db?

karlicoss:
Yeah, good suggestion! Just adding to context is a quick & easy workaround.

In general there are all kinds of metadata that could be useful, e.g:

- org-mode has tags
- reddit has users/subreddits/etc
- github (https://github.com/karlicoss/promnesia/issues/194) has tags/repos/users
- instant messaging is associated with a person
- tweets have authors

I need to think a bit more what's the best way to handle it.. because adding stuff to db is easy. The interesting question is what t do with it in the UI -- although just displaying it could be a good start. 
In the future would be interesting also to filter/search by metadata, but that's the hard part since involves frontend changes :)

ankostis:
Just made a PR #199 to include in Hypothesis rows a new context-line with (hash-)-tags;  the same can be done with github(labels), gmail(labels).

BUT when trying to search by a hash-prefixed tag e.g. `#freedom`, the extension takes too much time to respond, and this could be an escaping issue (and probably a security vulnerability?).

These are the logs when searching `freedom`:
```
[INFO    2021-02-16 13:36:42 promnesia.server server.py:270] /search freedom
[INFO    2021-02-16 13:36:42 promnesia.server server.py:164] url: freedom
[INFO    2021-02-16 13:36:42 promnesia.server server.py:167] normalised url: freedom
[DEBUG   2021-02-16 13:36:42 promnesia.server server.py:180] query: SELECT visits.norm_url, visits.orig_url, visits.dt, visits.locator_title, visits.locator_href, visits.src, visits.context, visits.duration 
    FROM visits 
    WHERE (visits.norm_url LIKE '%' || ? || '%' ESCAPE '/') OR (visits.orig_url LIKE '%' || ? || '%' ESCAPE '/') OR (visits.context LIKE '%' || ? || '%' ESCAPE '/') OR (visits.locator_title LIKE '%' || ? || '%' ESCAPE '/')
[DEBUG   2021-02-16 13:36:43 promnesia.server server.py:193] got 57 visits from db
[DEBUG   2021-02-16 13:36:43 promnesia.server server.py:204] responding with 57 visits
127.0.0.1 - - [16/Feb/2021 13:36:43] "POST /search HTTP/1.1" 200 30282
```
[...]

Issue #194, opened 2021-02-14
Missing `github` source?


ankostis:
From your [big-plan(TM)](https://beepb00p.xyz/myinfra_files/myinfra.svg), i see this data-flow: 

            .--(manual)-->GDPR-archive-->HPI.github.gdpr------.
    GitHub--+                                                 +-->promnesia.github(src)
            '--ghexport-->json-events--->HPI.github.ghexport--'

but promnesia is missing `promnesia.github(src)` module.
Is that on purpose?

karlicoss:
Oh thank you, nice catch! I guess I forgot to implement github module for promnesia!

PR #199, opened 2021-02-16
feat: store also Hypothes.is tags

ankostis:
(address #195) store Hypothesis tags in a new context line, prefixed by hash(`#`), separated by space.

### DIFF
--- a/doc/GUIDE.org
+++ b/doc/GUIDE.org
@@ -67,6 +67,7 @@ Searches in visits and their contexts. You can toggle search via:
 - hotkey: =Ctrl+Shift+H/Ctrl+Alt+H= (Chrome/Firefox correspondingly)
 - Promnesia context menu
 - sidebar button
+- search just for a =#tag=
 
 ** 'search around'
 Shows you browsing history 'around' another visit, useful to remember how/why did you get on the page.
--- a/src/promnesia/server.py
+++ b/src/promnesia/server.py
@@ -162,8 +162,10 @@ def search_common(url: str, where):
     config = EnvConfig.get()
 
     logger.info('url: %s', url)
-    original_url = url
-    url = canonify(url)
+    original_url = url and url.strip()
+    url = canonify(original_url)
+    if not url:  # Don't eliminate a "#tag" query.
+        url = original_url
     logger.info('normalised url: %s', url)
 
     visits0: List[Any] = []
--- a/src/promnesia/sources/hypothesis.py
+++ b/src/promnesia/sources/hypothesis.py
@@ -13,11 +13,14 @@ def index() -> Results:
             continue
         hl = h.highlight
         ann = h.annotation
+        tags = h.tags
         cparts = []
         if hl is not None:
             cparts.append(hl)
         if ann is not None:
             cparts.extend(['comment: ' + ann])
+        if tags:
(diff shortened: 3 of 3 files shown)


ankostis:
Unfortunately, when tags are stored as `#tag-text` searches fail, and all visits match (see [karlicoss/promnesia#195](https://github.com/karlicoss/promnesia/issues/195#issuecomment-779781446)).

Any ideas why? [**edit:** I have no experience with SQLAlchemy]

ankostis:
I did the change in server code discused in #194, but i'm not confident to build and install the extension, to test it, so i will leave the respective change to you @karlicoss.
You may merge this PR after i made some documentation changes for the new feature.

ankostis:
This PR is ready by me.

## Fixes required
Respective hanges to the JS-code of searching bookmarks & history are pending,
because currently it fetches all of them when searching for a `#hash-tag`.

## Enhancement for the future
Currently it matches also url-fragments in `context` e.g `#foo` matches `https://example.com#foo-bar`.

An improvement would be to execute this SQL with parameter `{tag}` when a hash-tag is queried:
```sql
SELECT * FROM visits WHERE context REGEXP '(?<!\w)#{tag}'
```

Also need to explicitly extract tags from *pocket*, *takeout* (labels), etc.

karlicoss:
Thanks! I guess it makes sense, will give it a think what would be a proper strategy for handling these.

> An improvement would be to execute this SQL with parameter {tag} when a hash-tag is queried:

Yep. But again, need to think first what's the most reasonable generic way to do this -- I want to avoid feature bloat :) 

> because currently it fetches all of them when searching for a #hash-tag.

Yeah. This is one of the most annoying bits to me as well, that the logic (especially normalization logic) is spread out between python and JS. At some point I was even considering automatic translation of the code, but it turned to be out more tedious than I expected. Maybe a good intermediate solution would be to rely on the backend for normalization & some other logic when it's present, and without the backed, fallback on the 'simple' normalization, at least for now.

If it's too spammy for you, you can disable local history & bookmarks in the settings for now, before it's properly fixed. (another nice feature for search to have -- to make it possible to specify which sources to search over)

P.S. CI jobs failed, but it's already fixed on master, so not a problem. Also need to finally figure out how to trigger github actions on an incoming pull request...

I suppose you're happy with running off a git version? I wanted to do a few other fixes and maybe release at the end off week.

ankostis:
> I suppose you're happy with running off a git version?

I do run all of your stuff like that, for editing on the sport in case bugs arise.
For instance i'm consistently bugged by issues when `twitter.all` & `github.all` don't have both backends (e.g. gdpr-archive & events) - they fail.

karlicoss:
Yeah, for twitter.all/github.all -- I guess need a more generic defensive mechanism (it's hard to make it dynamic enough & mypy-safe at the same time)

Although modyfying in place is perfectly valid way of configuring, you can also use overlays with your own `twitter/all.py` or `github/all.py` to exclude gdpr archives etc https://github.com/karlicoss/HPI/blob/master/doc/SETUP.org#addingmodifying-modules

You also might find https://github.com/karlicoss/HPI/issues/102 useful (and maybe if you have some thoughts on the subject, feel free to share!)

PR #202, opened 2021-02-19
Enable usage on NixOS

jakeisnt:
NixOS doesn't use the /bin directory, but still has `bash` in path. Ergo, to get this working properly with NixOS, the path has to be found from the environment.


karlicoss:
Thanks! good to get this in habit anyway, I do it for python, but keep forgetting for bash

PR #204, opened 2021-02-22
Add new source for Viber Desktop local sqlite db

ankostis:
# Add new source for Viber Desktop local sqlite db

Viber Desktop (at least in linux that i checked) stores its messages in a local SQLite file 
`~/.ViberPC/{user-phone-number}/viber.db` and it contains everything needed to devise Visits stream,
as shown in the the sample below:
image

The indexer collects:

- chat-name
- sender
- chat text
- title extracted from any previewed url
- chat-tags
- any URLs on the chat-text

The only one missing is a working `Locator.href`, which i couldn't think of anything other than:

    sqlite:///home/username/.ViberPC/123456789/viber.db?immutable=1#!Messages.EventId=3445` 

It is configured with:

```python
SOURCES = [
    Source(viber_desktop.index, "~/.ViberPC/your-phone-number/viber.db"),
]
```

* DOCs.have been updated;
* Sample config has been updated.


src/promnesia/sources/viber_desktop.py
```diff
@@ -0,0 +1,164 @@
...
+                    href=f"sqlite://{database}?immutable=1#!Messages.EventId={mid}",
+                ),
+            )
+
+    # TODO yield error if chatname or chat or smth else is null?
+    for row in db.query(query_str):
+        try:
+            yield from _handle_row(row)
+        except Exception as ex:
+            logger.warning(
```
karlicoss: Replying to your comment [here](https://github.com/karlicoss/promnesia/commit/4363b5797b30819b89a4e4908ea27b009b598112#commitcomment-47433572) 

Basically the options people usually use are:

1. throw immediately (as you suggested in the comment)
  This is problematic because then on any unexpected situation (like 1/10000 messages parsed incorrectly), the program will break. Considering we have to deal with closed protocols, formats change without warning, and it's annoying when your software breaks until you fix it, so we need to be more flexible and defensive here. In addition, imagine if one data source breaks the whole program -- sometimes it's justified (e.g. if you were working with important medical data or something like that), but not in case of this tool.

2. log and continue (as you're doing here)
   It works nice in terms of not breaking immediately. However then there is another danger -- imagine that for example Viber changes format for 'time' column to store the isoformatted string (instead of epoch), for whatever reason. Then promnesia indexer will carry on working (because of exception catching) -- it's good. However, you won't notice the error unless you regularly reviewing the logs, or maybe when you know for sure that Promnesia should fire but it doesn't.

So in Promnesia (and related projects that process data, like HPI) I'm doing something intermediate.

- catch the exceptions (where it makes sense, e.g. parsing)
- possibly log it if necessary
- but then `yield` it to the parent function, as a normal value (there is even an alias for [`Result`](https://github.com/karlicoss/promnesia/blob/9b7520feea9872e29520d466aa302b76ea068c52/src/promnesia/common.py#L128-L130) )
 
This allows the top level handler to exit with code 1 (so you get a cron email/systemd job failure etc and notice the error), while still doing what's the program supposed to do (process as much as it can and insert in the database).

As more examples why it's useful in Promnesia, errors are also inserted in the database:
- this makes it easier to test in some cases (instead of intercepting logs, just check the database)
- later it will allow to overview the errors form the browser which would be much more convenient for most users than messing with systemd logs or whatnote

I write more about it [here](https://beepb00p.xyz/mypy-error-handling.html), you might enjoy it!

src/promnesia/sources/viber_desktop.py
```diff
@@ -0,0 +1,164 @@
...
+    def _handle_row(row) -> Results:
+        text = row["text"]
+        assert (
+            text
+        ), f"sql-query should have eliminated messages not containing 'http': {text}"
+        urls = extract_urls(text)
+        if not urls:
+            return
+        dt = from_epoch(row["time"] // 1000)  # timestamps are stored x100 this db
+        mid: str = unwrap(row["mid"])
```
karlicoss: Oh! my bad, for `unwrap` in telegram module I intented to do `notnone` instead, [like in HPI](https://github.com/karlicoss/HPI/blob/ad924ebca84b0846c98401b10e61488c45cf1e9c/my/core/error.py#L18-L20). The reason was to make sure null rows aren't converted to `None` in python and avoid downstream errors.

`Exception` can never be in the sqlite row, so you can either remove `unwrap` from this module, or use `notnone` instead, up to you.

src/promnesia/sources/viber_desktop.py
```diff
@@ -0,0 +1,164 @@
...
+
+    # TODO for displaying maybe better not to expand/absolute, 
+    # but it's safer for debugging resolved.
+    path = Path(database).expanduser().resolve().absolute()
+    assert path.is_file(), path
+
+    # TODO context manager?
+    db = dataset_readonly(path)  # TODO could check is_file inside
+
+    query_str = textwrap.dedent(
```
karlicoss: nice, didn't know about `dedent` (and the whole textwrap module!)

src/promnesia/sources/viber_desktop.py
```diff
@@ -0,0 +1,164 @@
...
+import textwrap
+from pathlib import Path
+from typing import Optional, TypeVar, Union
+from urllib.parse import \
+    unquote  # TODO mm, make it easier to rememember to use...
+
+from ..common import (Loc, PathIsh, Results, Visit, echain, extract_urls,
+                      from_epoch, get_logger)
+
+# TODO potentially, belongs to my. package
```
karlicoss: Ideally this module (i.e. the query/parsing part) would be part of HPI (so viber data can be used by other apps).
The only reason it's not is that the promnesia module existed before I started HPI project :)
But totally ok to keep here for now, can always refactor later if needed.

src/promnesia/sources/viber_desktop.py
```diff
@@ -0,0 +1,164 @@
...
+    # see https://github.com/pudo/dataset/issues/136#issuecomment-128693122
+    import sqlite3
+
+    import dataset  # type: ignore
+
+    creator = lambda: sqlite3.connect(f"file:{db}?immutable=1", uri=True)
+    return dataset.connect("sqlite:///", engine_kwargs={"creator": creator})
+
+
+def index(database: PathIsh) -> Results:
```
karlicoss: One nice thing to do could be to make it optional , and it it's None, then do glob over `"~/.ViberPC/*/viber.db"`, then the user could just specify `Source(viber_desktop.index)` in the config without extra arguments.
But up to you.

src/promnesia/sources/viber_desktop.py
```diff
@@ -0,0 +1,164 @@
...
+
+        url_title = _parse_json_title(infojson)
+        if url_title:
+            text = f"title: {url_title}\n\n{text}"
+
+        for u in urls:
+            # https://www.reddit.com/r/Telegram/comments/6ufwi3/link_to_a_specific_message_in_a_channel_possible/
+            # hmm, only seems to work on mobile app, but better than nothing...
+            yield Visit(
+                url=unquote(u),
```
karlicoss: are they quoted in the viber database? Might be unnecessary (and if they are do you mind adding a comment?)

src/promnesia/sources/viber_desktop.py
```diff
@@ -0,0 +1,164 @@
...
+                dt=dt,
+                context=text,
+                locator=Loc.make(
+                    title=f"chat({mid}) from {sender}@{chatname}",
+                    href=f"sqlite://{database}?immutable=1#!Messages.EventId={mid}",
```
karlicoss: interesting! does it actually work, i.e. if you press on the link it can jump to this message in sqlitebrowser? (or what are you using?)

src/promnesia/sources/viber_desktop.py
```diff
@@ -0,0 +1,164 @@
...
+            tags = "".join(f"#{t}" for t in tags.split())
+            text = f"{text}\n\n{tags}"
+
+        url_title = _parse_json_title(infojson)
+        if url_title:
+            text = f"title: {url_title}\n\n{text}"
+
+        for u in urls:
+            # https://www.reddit.com/r/Telegram/comments/6ufwi3/link_to_a_specific_message_in_a_channel_possible/
+            # hmm, only seems to work on mobile app, but better than nothing...
```
karlicoss: can remove this comment I guess

doc/SOURCES.org
```diff
@@ -136,6 +136,11 @@ for importer, name, ispkg in sorted(pkgutil.walk_packages(
 
   Uses [[https://github.com/fabianonline/telegram_backup#readme][telegram_backup]] database for messages data
 
+- [[file:../src/promnesia/sources/viber_desktop.py][promnesia.sources.viber_desktop]]
+
+  Uses the local Viber Desktop SQLite databse in your home directory,
```
karlicoss: typo in 'database'

karlicoss (changes requested):
Very nice, thanks so much, especially for figuring out the query! 
Looks good, there are some inor comments (and some optional, so up to you if you want to fix them, otherwise I can clean up sometime later).

I don't have Viber so can't test it (to ensure it doesn't break later). Do you know by any chance, if there some public viber database somewhere on github, so it would be possible to run some basic tests against it? Otherwise not a big problem, hopefully the most potentially fragile part (query) won't need to change at all for a while.

src/promnesia/sources/viber_desktop.py
ankostis: I was having this question :-)

src/promnesia/sources/viber_desktop.py
ankostis: Resolved in 4c2c5d6.

src/promnesia/sources/viber_desktop.py
ankostis: Ofcourse not :-)
 
But i needed something to contain internal information to help the user locate the message, besides the chat-name & the date, already visible in the sidebar.  And i wanted to be clear that it requires _programmatic-access_, hence the `sqlite` url-schema, which can be used verbatim as a connection URI.
I think the fragment with the `table.column=<mid>` criterio would not matter (but neither work), but haven't checked it.

Is there some other idea (like telegram forwarding- url service)?

src/promnesia/sources/viber_desktop.py
ankostis: No they are not, removed it.
Resolved in f4553ed.

ankostis (commented):
>  Do you know by any chance, if there some public viber database somewhere on github

No.
I could empty mine quasi-completely, to keep just some for running the query, and upload it here or the creation sql-scripts.
Even better if i could upload the DDL+insert statements it in the sources, to create test-case in the sources.

Besides this requiring some time-investment from my side, it leads to another question:
maybe se shpould move the query part of the code in HPI, no?

src/promnesia/sources/viber_desktop.py
ankostis: Nice idea!
[edit] actually `SOURCES=[..., viber]` would suffice.

src/promnesia/sources/viber_desktop.py
ankostis: One day i may submit a mass-update PR for all sources needing `dedent` affection :-)

src/promnesia/sources/viber_desktop.py
ankostis: (read it before my own suggestion on that, [below](https://github.com/karlicoss/promnesia/pull/204#pullrequestreview-597263776)).
Yes i agree.
Will transplant part of this PR in HPI.

I will wait for your review to finish here, before spreading stuff in 2 projects.

src/promnesia/sources/viber_desktop.py
ankostis: Resolved in a8c6dd0.

src/promnesia/sources/viber_desktop.py
ankostis: I managed to open the Viber DB in `sqlitebrowser` in linux, 
by ifirst making a "hard" link to the db-file, replacing the suffix from `.db --> .sqlite`,
an then simply referring to it with `file:///...sqlite` uri.

Would that be ok (for `promnesia index` to make a hard-link in huser's viber-config directory??

ankostis: I implemented the `file:://...sqlite` locator-URIs in df948c3.

Unfortunately, the links do not work, due to browser's security preventing opening local file-links, i guess,
because when i copy-past the link in another tab, it does offer me to open it with `sqlitebrowser`.

src/promnesia/sources/viber_desktop.py
ankostis: Implemented in 2d5c3c4.

doc/SOURCES.org
ankostis: Fixed in 2d5c3c4.

karlicoss:
Posting as a separate comment so it's not buried in the code comments; might be useful for someone else later.

So, neither symlink/hardlink for `.db` -> `.sqlite` file worked for me so I've gone down the rabbithole trying to debug it...
I checked mime types, and seems all good (same for .sqlite extension)

    $ file karlicoss.db 
    karlicoss.db: SQLite 3.x database, last written using SQLite version 3027002
    $ mimetype karlicoss.db
    karlicoss.db: application/vnd.sqlite3
    $ mimeopen karlicoss.db 
    Opening "karlicoss.db" with DB Browser for SQLite  (application/vnd.sqlite3


Ok, but MIME not only examines the extension, but also looks at the 'magic bytes' in the file (that would explain how it can guess '.db'). But browser can only guess by the extension type, it can't look in the file before downloading... let's check how the extension guessing works...

    $ mimetype does-not-exist.pdf
    does-not-exist.pdf: application/pdf # OK 
    $ mimetype does-not-exist.sqlite
    does-not-exist.sqlite: 
    $ mimetype does-not-exist.db
    does-not-exist.db:

Aha! So the system doesn't actually know `.sqlite` (let alone `.db`) extensions.
So what can we do? 

Some search https://stackoverflow.com/a/31836/706389 suggests that you need to prepare an xml and install it with `xdg-mime` (at least if you're on Linux).
But wonder what are the default sqlite file extensions it expects?

    $ rg -s 'pattern.*sqlite' /usr/share/mime                                                                                                                                                                                                  20:35:17
    /usr/share/mime/application/vnd.sqlite3.xml
    54:  <glob pattern="*.sqlite3"/>
    
    /usr/share/mime/application/x-sqlite2.xml
    54:  <glob pattern="*.sqlite2"/>
    
    /usr/share/mime/packages/freedesktop.org.xml
    2105:    <glob pattern="*.sqlite2"/>
    2161:    <glob pattern="*.sqlite3"/>

So here we go -- apparently it knows .sqlite2/.sqlite3, but not .sqlite.

Ok, let's prepare the xml...

    $ cat sqlite-mime.xml
    <?xml version="1.0"?>
    <mime-info xmlns='http://www.freedesktop.org/standards/shared-mime-info'>
      <mime-type type="application/sqlite-extra">
        <comment>Extra sqlite extensions</comment>
        <glob pattern="*.db"/>
        <glob pattern="*.sqlite"/>
      </mime-type>
    </mime-info>

    $ xdg-mime install sqlite-mime.xml && update-desktop-database

And...

    $ mimetype does-not-exist.db
    does-not-exist.db: application/sqlite-extra

Win! Now we only need to associate sqlitebrowser with it:

    $ xdg-mime default sqlitebrowser.desktop application/sqlite-extra

And after that Firefox finally offers to open the database!

karlicoss:
Also while debugging the above, found another way, posting it here since might be useful somewhere else later...

Found some help on changing the application handler in Firefox, but unclear how to add a new format: https://support.mozilla.org/en-US/kb/applications-panel-set-how-firefox-handles-files

Following advice from here: https://support.mozilla.org/en-US/questions/1235051#answer-1158433

    $ ~/.mozilla/firefox/$PROFILEDIR/handlers.json

    "mimeTypes": {
        "application/pdf": {
            "action": 3,
            "extensions": [
                "pdf"
            ]
        },
        "application/vnd.sqlite3": {
            "action": 4,
            "extensions": [
                "sqlite",
                "db"
            ],
            "handlers": [
                {
                    "name": "sqlitebrowser",
                    "path": "/usr/bin/sqlitebrowser"
                }
            ]
        }


Explanation for action field from here: https://support.mozilla.org/fr/questions/1262664#answer-1232236

    saveToDisk = 0;
    alwaysAsk = 1;
    useHelperApp = 2;
    handleInternally = 3;
    useSystemDefault = 4;


After reloading Firefox this way also works. But it's a bit sad the functionality isn't exposed to the GUI :(

Also, neither of the ways work in Chrome -- it seems to have a mind of its own, and quick search seems to suggest there is no way without custom extensions  :vomiting_face:

src/promnesia/sources/viber_desktop.py
karlicoss: > I think the fragment with the table.column=<mid> criterio would not matter

Yeah fair enough. Just asked in case it was some hidden feature, it could be convenient :) 

I think the closest possible thing that sqlitebrowser has is `--sql` argument (although I couldn't get it to work), but it wouldn't work file `file://` handler unfortunately.

> Is there some other idea (like telegram forwarding- url service)?

Quick search gives something: https://stackoverflow.com/a/41324451/706389, but it's probably unlikely to work on desktop unfortunately :( Such things are always better on mobile.

> Would that be ok (for promnesia index to make a hard-link in huser's viber-config directory??

Hmm. Hardlinks are pretty extreme and opaque IMO -- the user might damage the original database without knowing it. But I see the problem with symlinks, they seem to be resolved before opening and it then tries to open '.db' file.
Maybe comment the hardlink logic for now? I'm worried that someone trashes their database; maybe we can come up with a more robust idea later? 
Also see the comment https://github.com/karlicoss/promnesia/pull/204#issuecomment-785374009 how to handle it on user site without hardlinking (a little tedious, but you only need to do it once, and maybe later we can integrate these instructions in promnesia, similar to [open-with-editor(https://github.com/karlicoss/open-in-editor) ?)

karlicoss:
> No. I could empty mine quasi-completely, to keep just some for running the query, and upload it here or the creation sql-scripts. Even better if i could upload the DDL+insert statements it in the sources, to create test-case in the sources.

Up to you -- don't expect anyone to break it, and I don't mind if we add tests for this later. I know how tedious and time consuming this 'database emptying' can be.
For uploading, I usually use submodules (e.g. https://github.com/karlicoss/HPI/blob/ad924ebca84b0846c98401b10e61488c45cf1e9c/.gitmodules#L1-L3) it keep the main repository lean (since database binaries are only needed in tests anyway).

> Besides this requiring some time-investment from my side, it leads to another question: maybe se shpould move the query part of the code in HPI, no?

Yep, could do, e.g. iterator over messages could be useful, and then it would be imported in Promnesia for URL extraction.
But also similar -- I'm happy to merge it here, and later you can move it over to HPI -- up to you!

karlicoss:
Looks good to me! 🎉 Let me know what you think about the hardlink thing, and after that I'm happy to merge if you are.

ankostis:
>  Let me know what you think about the hardlink thing, and after that I'm happy to merge if you are.

I will remove the had-link (your xdg-handler fits by purpose), and then you can merge it.
Tests will come later (submodules is the only thing in git that i'm not fluent).

src/promnesia/sources/viber_desktop.py
ankostis: Great research.
Will drop the hard-link part.  Closing this review.

ankostis:
Removed hard-link hack in 264f543,
Read to be merged.

Tell me when to split it in HPI, better in 2 separate PRs, no?

karlicoss:
Thank you! :tada: merged

For two PRs -- you mean one would be to HPI (e.g. `my.viber.local` module), another to Promnesia to actually switch to use `my.viber.local` data? Yeah that makes sense

ankostis:
> Ok, let's prepare the xml...

Many thanks!

Would it make sense to augment [karlicoss/open-in-editor](https://github.com/karlicoss/open-in-editor) to install those handlers you crafted in those 2 messages?

karlicoss:
Not sure about adding this to open-in-editor.. I guess it's somewhat unrelated. 
I guess I won't mind it as a separate script in `promnesia/script` or `open-in-editor` repository perhaps? And maybe later it will be clearer where all this stuff actually belongs?

ankostis:
Will you add it?
It's really usefull since in just the mimetype registration made my gnome offer me a dialog-choice for which app to use to view aqlite files, and after choosing `sqlitebrowser` it works thereafter.

karlicoss:
Ah, great to know that even mimetype is enough! Yep, will add later.

PR #208, opened 2021-02-27
Fix schema in viber locator hrefs from file-->editor

ankostis:
In promnesia's search-results tab, links like `file:///.ViberPC/123/viber.db` do not work,
due to security (i guess, bc these urls work fine when pasted in a new tab).

The custom schema `editor:` offers a better UX, 
since it gives the user the choice of which app to use to open the Viber's db-file (e.g. sqlitebrowser`),
when the mime-type is known to the OS (see the 1st resource, below).

These resources are relevant:
* [install a `sqlite` mime-type handler](https://github.com/karlicoss/promnesia/pull/204#issuecomment-785374009)
* [change *firefox* handler](https://github.com/karlicoss/promnesia/pull/204#issuecomment-785376783)
* [karlicoss/open-in-editor](https://github.com/karlicoss/open-in-editor) custom handler

@karlicoss I want your tip about the mime-type handler to be included in the Viber docs;  any idea where or how to include it?

### DIFF
--- a/src/promnesia/sources/viber.py
+++ b/src/promnesia/sources/viber.py
@@ -82,6 +82,7 @@ def messages_query() -> str:
         LEFT JOIN Groups AS G
             ON E.ChatId = G.ChatId
         WHERE
+            M.ClientFlag != 64  AND             /* edited messages */
             text LIKE '%http%'
         ORDER BY time;
         """
--- a/src/promnesia/sources/viber.py
+++ b/src/promnesia/sources/viber.py
@@ -96,7 +96,7 @@ def _parse_json_title(js) -> str:
             return js.get("Title")
 
 
-def _handle_row(row: dict, db_path: PathLike) -> Results:
+def _handle_row(row: dict, db_path: PathLike, locator_schema: str) -> Results:
     text = row["text"]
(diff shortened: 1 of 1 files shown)


ankostis:
This PR fixes also a minor bug in the SQL-query, which previously fetched even overwritten due-to-edits viber messages.

karlicoss:
Another way to achieve this is the post-index hook https://github.com/karlicoss/promnesia/blob/aded41c5271c57d704f721d48670ee5135cc5f7c/tests/indexer_test.py#L284-L303
So e.g. you could do something like "if source is viber, then replace url etc etc". 

But if you think it makes sense as a setting too, happy to merge. 
I guess the whole thing with mime type configuration needs a bit more thinking. Maybe in general Promnesia could even offer multiple ways of opening the link (e.g. via sqlitebrowser **and** via web app (if it's possible, e.g. like for Telegram), etc, etc). Kind of related to https://github.com/karlicoss/promnesia/pull/147#issuecomment-718293549

ankostis:
I think a better idea would be to make the url more configurable, e.g. store a print-pattern in the config,
but i believe we can merge this for the forseable future.

PR #209, opened 2021-02-28
Added ignore kwarg to traverse() in common, allowing directory pruning

Cobertos:
This adds an optional kwarg to traverse() in common.py, allowing directories that are ignored to be pruned by fdfind and find, instead of iterating over and excluding later.

There's also two tests to make sure that this is working with both fdfind and find. Passes on my machine

### DIFF
--- a/src/promnesia/common.py
+++ b/src/promnesia/common.py
@@ -443,8 +443,12 @@ def traverse(root: Path, *, follow: bool=True, ignore: List[str]=[]) -> Iterable
     if _is_windows:
         # on windows could use 'forfiles'... but probably easier not to bother for now
         # todo coild use followlinks=True? walk could end up in infinite loop?    
-        for r, _, files in os.walk(root):
-            yield from (Path(r) / f for f in files)
+        for r, dirs, files in os.walk(root):
+            # Remove dirs specified in ignore (clone dirs() as we have to remove in place)
+            for i, d in enumerate(list(dirs)):
+                if d in ignore:
+                    del dirs[i]
+            yield from (Path(r) / f for f in files if f not in ignore)
         return
 
     from .compat import Popen, PIPE
--- a/src/promnesia/common.py
+++ b/src/promnesia/common.py
@@ -5,6 +5,7 @@ from typing import NamedTuple, Set, Iterable, Dict, TypeVar, Callable, List, Opt
 from pathlib import Path
 from glob import glob
 import itertools
+from more_itertools import intersperse
 import logging
 from functools import lru_cache
 import shutil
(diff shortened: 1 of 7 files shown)


Cobertos:
Hmm, I forgot about the Windows implementation

karlicoss:
Thanks, ill review later when i get to my computer.
I think there are only a few Windows users, so unless you have easy access
to windows computer, no need to worry for now! (as long as it doesn't crash)

Cobertos:
Oh okay, I already fixed x3

Cobertos:
Oh, I also forgot to mention that this requires `more_itertools` for the `disperse` function. I did not add the dependency, let me know the best way to handle that, considering all the different sort ways there are to install.

karlicoss:
Ah, it's already a dependency: https://github.com/karlicoss/promnesia/blob/20a07c23d150e2206f22d1afa5b1a4f1a17e3305/setup.py#L34

so no problem with that. `more_itertools` is valuable enough to have as dependency (plus it's one of the most common dependencies anyway, so usually already installed)

karlicoss:
Sorry, need to figure out how to run github CI against other's people's PRs -- at the moment it's only triggered by my own pushes only, so I need to checkout PR and push it to trigger.

-----

Looks like it's failing `ignore_find` test (`CI=true tox -e tests -- -k ignore_find`)

```
E         Full diff:
E           [
E         +  PosixPath('/adhoc-project/tests/testdata/traverse/imhere.txt'),
E            PosixPath('/adhoc-project/tests/testdata/traverse/imhere2/real.txt'),
E         -  PosixPath('/adhoc-project/tests/testdata/traverse/imhere.txt'),
E           ]
```
guess find doesn't guarantee order. 

-------------

mypy also is going to complain `CI=true tox -e mypy-core`:

```
src/promnesia/common.py: note: In function "find_args":
src/promnesia/common.py:401: error: Argument 1 to "chain" has incompatible type "*Iterator[List[str]]"; expected
"Iterable[List[str]]"  [arg-type]
            ignore_names = list(itertools.chain(*intersperse(['-o'], ignore_names)))
                                                 ^
src/promnesia/common.py:411: error: List item 2 has incompatible type "List[Sequence[str]]"; expected "str"  [list-item]
            *prune_dir_args,
             ^
src/promnesia/common.py:413: error: List item 5 has incompatible type "List[Sequence[str]]"; expected "str"  [list-item]
            *ignore_file_args
             ^
src/promnesia/common.py: note: In function "fdfind_args":
src/promnesia/common.py:425: error: Argument 1 to "chain" has incompatible type "*List[List[str]]"; expected
"Iterable[List[str]]"  [arg-type]
            ignore_args = list(itertools.chain(*ignore_args))
                                                ^
src/promnesia/common.py:429: error: List item 1 has incompatible type "List[List[str]]"; expected "str"  [list-item]
            *ignore_args,
             ^
```

If you allow edits from maintainers, I don't mind quickly fixing it.

Cobertos:
Hmm, it should allow maintainers to edit. The checkbox was checked?

Oh yeah, the order, I usually use pythons unittest asserts. Probably just
needs a Set comparison

What is the Python typing stuff, how do I run it? Never seen that before.

karlicoss:
Ah, indeed, pushing onto the branch works, not sure why it didn't first time I tried!

Python typing stuff is [mypy](https://github.com/python/mypy). In principle, possible to run it as `mypy -p promnesia`, but due to third party modules and other complexities, the most reliable is to do it via tox (e.g. `CI=true tox -e mypy-core`) 

https://github.com/karlicoss/promnesia/blob/708d0739c2f3112b03370070af2a8d35d5642983/tox.ini#L32-L39

Either way, it was nothing major, just mypy being a bit picky about reusing the name with a different type, I quickly fixed it.

Cobertos:
Oh I see, I haven't used tox at all. I will have to look into how that's all set. Looks like you got the PR CI stuff working too.

karlicoss:
Thanks for the contribution!

PR #211, opened 2021-02-28
FEAT(index) add --sources & --overwrite to CLI to partially update db from config sources

### LINKED ISSUE #20: real time indexing

ankostis:
Fixes #20
 
### new `--sources` option(s)
* args can be both Source.name & Source's positional index.
* useful with dynamic *partial* updates, below (now the default), so that the original config file ca be used.

### new `--overwrite` option
- Added a single CLI option almost-described in the 2nd [alternative case](#issuecomment-787917633) explained below,
  due to simplicity.  if missing, defaults to "dynamic partial updates" as described in [this](#issuecomment-788375379) and [this](#issuecomment-788387359).
- Dropped `PROMNESIA_INDEX_POLICY` env-var, since it is the default behavior now.
- Function defaults for `--overwrite` are `False`, as suggested in [the comment below](#issuecomment-778382929).
- Both `promnesia index` & `promnesia demo` sub-commands have been updated.
- All update/overwrite decision logic moved to `promnesia/__main__.py`.:

### Various
- Fix: create db-output directory in advance.

### DIFF
--- a/src/promnesia/__main__.py
+++ b/src/promnesia/__main__.py
@@ -131,6 +131,7 @@ def do_demo(
         params: Sequence[str],
         port: Optional[str],
         config_file: Optional[Path],
+        dry: bool=False,
         name: str='demo',
         sources_subset: Iterable[Union[str, int]]=(),
         overwrite_db: bool=False,
@@ -150,7 +151,7 @@ def do_demo(
             )
             config.instance = cfg
 
-        errors = list(_do_index(sources_subset=sources_subset, overwrite_db=overwrite_db))
+        errors = list(_do_index(dry=dry, sources_subset=sources_subset, overwrite_db=overwrite_db))
         if len(errors) > 0:
             logger.error('%d errors during indexing (see logs above for backtraces)', len(errors))
         for e in errors:
@@ -286,30 +287,37 @@ def _ordinal_or_name(s: str) -> Union[str, int]:
 def main() -> None:
     # TODO longer, literate description?
 
+    def add_index_args(parser: argparse.ArgumentParser, default_config_path: PathIsh=None) -> None:
+        """
+        :param default_config_path:
+            if not given, all :func:`demo_sources()` are run
+        """
+        parser.add_argument('--config', type=Path, default=default_config_path, help='Config path')
+        parser.add_argument('--dry', action='store_true', help="Dry run, won't touch the database, only print the results out")
+        parser.add_argument(
+            '--sources',
+            required=False,
+            action="extend",
+            nargs="+",
+            type=_ordinal_or_name,
+            metavar="SOURCE",
+            help="Source names (or their 0-indexed position) to index.",
+        )
+        parser.add_argument(
+            '--overwrite',
+            required=False,
+            action="store_true",
+            help="Empty db before populating it with newly indexed visits."
+            "  If interrupted, db is left untouched."
+        )
+
     F = lambda prog: argparse.ArgumentDefaultsHelpFormatter(prog, width=120)
     p = argparse.ArgumentParser(formatter_class=F) # type: ignore
     subp = p.add_subparsers(dest='mode', )
     ep = subp.add_parser('index', help='Create/update the link database', formatter_class=F)
-    ep.add_argument('--config', type=Path, default=default_config_path(), help='Config path')
-    ep.add_argument('--dry', action='store_true', help="Dry run, won't touch the database, only print the results out")
+    add_index_args(ep, default_config_path())
     # TODO use some way to override or provide config only via cmdline?
     ep.add_argument('--intermediate', required=False, help="Used for development, you don't need it")
-    ep.add_argument(
-        '--sources',
-        required=False,
-        action="extend",
-        nargs="+",
-        type=_ordinal_or_name,
-        metavar="SOURCE",
-        help="Source names (or their 0-indexed position) to index.",
-    )
-    ep.add_argument(
-        '--overwrite',
-        required=False,
-        action="store_true",
-        help="Empty db before populating it with newly indexed visits."
-        "  If interrupted, db is left untouched."
-    )
 
     sp = subp.add_parser('serve', help='Serve a link database', formatter_class=F) # type: ignore
     server.setup_parser(sp)
@@ -323,29 +331,13 @@ def main() -> None:
     ap.add_argument('--name', type=str, default='demo'               , help='Set custom source name')
     add_port_arg(ap)
     ap.add_argument('--no-serve', action='store_const', const=None, dest='port', help='Pass to only index without running server')
-    ap.add_argument('--config', type=Path, required=False            , help='Config to run against. If omitted, will use empty base config')
     ap.add_argument(
         '--as',
         choices=list(sorted(demo_sources().keys())),
         default='guess',
         help='Promnesia source to index as (see https://github.com/karlicoss/promnesia/tree/master/src/promnesia/sources for the full list)',
     )
-    ap.add_argument(
-        '--sources',
-        required=False,
-        action="extend",
-        nargs="+",
-        type=_ordinal_or_name,
-        metavar="SOURCE",
-        help="Source names (or their 0-indexed position) to index.",
-    )
-    ap.add_argument(
-        '--overwrite',
-        required=False,
-        action="store_true",
-        help="Empty db before populating it with newly indexed visits."
-        "  If interrupted, db is left untouched."
-    )
+    add_index_args(ap)
     ap.add_argument('params', nargs='*', help='Optional extra params for the indexer')
 
     isp = subp.add_parser('install-server', help='Install server as a systemd service (for autostart)', formatter_class=F)
@@ -403,6 +395,7 @@ def main() -> None:
                 params=args.params,
                 port=args.port,
                 config_file=args.config,
+                dry=args.dry,
                 name=args.name,
                 sources_subset=args.sources,
                 overwrite_db=args.overwrite,
--- a/src/promnesia/__main__.py
+++ b/src/promnesia/__main__.py
@@ -5,7 +5,7 @@ import sys
 from typing import List, Tuple, Optional, Dict, Sequence, Iterable, Iterator, Union
 from pathlib import Path
 from datetime import datetime
-from .compat import check_call
+from .compat import check_call, register_argparse_extend_action_in_pre_py38
 from tempfile import TemporaryDirectory
 
 
@@ -292,6 +292,7 @@ def main() -> None:
         :param default_config_path:
             if not given, all :func:`demo_sources()` are run
         """
+        register_argparse_extend_action_in_pre_py38(parser)
         parser.add_argument('--config', type=Path, default=default_config_path, help='Config path')
         parser.add_argument('--dry', action='store_true', help="Dry run, won't touch the database, only print the results out")
         parser.add_argument(
--- a/src/promnesia/__main__.py
+++ b/src/promnesia/__main__.py
@@ -1,5 +1,4 @@
 import argparse
-import os
 import logging
 import inspect
 import sys
@@ -63,7 +62,7 @@ def iter_all_visits(sources_subset: Iterable[Union[str, int]]=()) -> Iterator[Re
         logger.warning("unknown --sources: %s", ", ".join(repr(i) for i in sources_subset))
 
 
-def _do_index(dry: bool=False, sources_subset: Iterable[Union[str, int]]=()) -> Iterable[Exception]:
+def _do_index(dry: bool=False, sources_subset: Iterable[Union[str, int]]=(), overwrite_db=False) -> Iterable[Exception]:
     # also keep & return errors for further display
     errors: List[Exception] = []
     def it() -> Iterable[Res[DbVisit]]:
@@ -78,7 +77,7 @@ def _do_index(dry: bool=False, sources_subset: Iterable[Union[str, int]]=()) -> 
         for v in res:
             print(v)
     else:
-        dump_errors = visits_to_sqlite(it())
+        dump_errors = visits_to_sqlite(it(), overwrite_db=overwrite_db)
         for e in dump_errors:
             logger.exception(e)
             errors.append(e)
@@ -89,10 +88,11 @@ def do_index(
         config_file: Path,
         dry: bool=False,
         sources_subset: Iterable[Union[str, int]]=(),
+        overwrite_db: bool=False,
     ) -> None:
     config.load_from(config_file) # meh.. should be cleaner
     try:
-        errors = list(_do_index(dry=dry, sources_subset=sources_subset))
+        errors = list(_do_index(dry=dry, sources_subset=sources_subset, overwrite_db=overwrite_db))
     finally:
         config.reset()
     if len(errors) > 0:
@@ -133,6 +133,7 @@ def do_demo(
         config_file: Optional[Path],
         name: str='demo',
         sources_subset: Iterable[Union[str, int]]=(),
+        overwrite_db: bool=False,
     ) -> None:
     from pprint import pprint
     with TemporaryDirectory() as tdir:
@@ -149,7 +150,7 @@ def do_demo(
             )
             config.instance = cfg
 
-        errors = list(_do_index(sources_subset=sources_subset))
+        errors = list(_do_index(sources_subset=sources_subset, overwrite_db=overwrite_db))
         if len(errors) > 0:
             logger.error('%d errors during indexing (see logs above for backtraces)', len(errors))
         for e in errors:
@@ -300,8 +301,14 @@ def main() -> None:
         nargs="+",
         type=_ordinal_or_name,
         metavar="SOURCE",
-        help="Source names (or their 0-indexed position) to index."
-        "  If missing, db is recreated empty and all sources are indexed.",
+        help="Source names (or their 0-indexed position) to index.",
+    )
+    ep.add_argument(
+        '--overwrite',
+        required=False,
+        action="store_true",
+        help="Empty db before populating it with newly indexed visits."
+        "  If interrupted, db is left untouched."
     )
 
     sp = subp.add_parser('serve', help='Serve a link database', formatter_class=F) # type: ignore
@@ -323,8 +330,22 @@ def main() -> None:
         default='guess',
         help='Promnesia source to index as (see https://github.com/karlicoss/promnesia/tree/master/src/promnesia/sources for the full list)',
     )
-    ap.add_argument('--sources', required=False, action="extend", nargs="+", type=_ordinal_or_name,
-                    help="Subset of source(s) to run (name or 0-indexed position);  use `promnisia --dry` to view sources")
+    ap.add_argument(
+        '--sources',
+        required=False,
+        action="extend",
+        nargs="+",
+        type=_ordinal_or_name,
+        metavar="SOURCE",
+        help="Source names (or their 0-indexed position) to index.",
+    )
+    ap.add_argument(
+        '--overwrite',
+        required=False,
+        action="store_true",
+        help="Empty db before populating it with newly indexed visits."
+        "  If interrupted, db is left untouched."
+    )
     ap.add_argument('params', nargs='*', help='Optional extra params for the indexer')
 
     isp = subp.add_parser('install-server', help='Install server as a systemd service (for autostart)', formatter_class=F)
@@ -358,13 +379,20 @@ def main() -> None:
         p.print_help(sys.stderr)
         sys.exit(1)
 
+    logger.info("CLI args: %s", args)
+
     # TODO maybe, it's better for server to compute intermediate represetnation?
     # the only downside is storage. dunno.
     # worst case -- could use database?
 
     with get_tmpdir() as tdir: # TODO??
         if args.mode == 'index':
-             do_index(config_file=args.config, dry=args.dry, sources_subset=args.sources)
+            do_index(
+                config_file=args.config,
+                dry=args.dry,
+                sources_subset=args.sources,
+                overwrite_db=args.overwrite,
+            )
         elif args.mode == 'serve':
             server.run(args)
         elif args.mode == 'demo':
@@ -377,6 +405,7 @@ def main() -> None:
                 config_file=args.config,
                 name=args.name,
                 sources_subset=args.sources,
+                overwrite_db=args.overwrite,
                 )
         elif args.mode == 'install-server': # todo rename to 'autostart' or something?
             install_server.install(args)
--- a/src/promnesia/__main__.py
+++ b/src/promnesia/__main__.py
@@ -3,7 +3,7 @@ import os
 import logging
 import inspect
 import sys
-from typing import List, Tuple, Optional, Dict, Sequence, Iterable, Iterator
+from typing import List, Tuple, Optional, Dict, Sequence, Iterable, Iterator, Union
 from pathlib import Path
 from datetime import datetime
 from .compat import check_call
@@ -19,7 +19,7 @@ from .dump import visits_to_sqlite
 from .extract import extract_visits, make_filter
 
 
-def iter_all_visits() -> Iterator[Res[DbVisit]]:
+def iter_all_visits(sources_subset: Iterable[Union[str, int]]=()) -> Iterator[Res[DbVisit]]:
     cfg = config.get()
     output_dir = cfg.output_dir
     # not sure if belongs here??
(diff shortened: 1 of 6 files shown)


ankostis:
I'd like to add here more commits, to implement a new CLI option <del>`--update`</del> `--overwrite <policy>` from #20, where `<policy>` can be:

- `yes`: delete & recreate promnesia db with newly indexed visits
- `no`: keep existing visits and merge new ones collected
- `if_partial`: (default when option is missing) `update` when `--sources` has been given, `overwrite` otherwise

Also documentation is missing (and a `CHANGES.md` would be good to have file).

ankostis:
Another alternative would be to add just 2 mutual-exclusive options: `--update` & `--overwrite`, which would be terser for users.  But the previous syntax is more extendable, if in the future a new policy appears.

If will implement this, and await feedback.

ankostis:
Implemented the 2nd alternative (and updated OP above with the new feature).

ankostis:
This PR is now only missing the _docs_ & _examples_, and at the moment has taken the following design choices:

### Added `--update` / `--overwrite` mutual exclusive options
- Added the CLI options described in the 2nd [alternative case](#issuecomment-787917633) explained below,
  due to simplicity.
- The precedence for deciding whether to _update_ or _overwrite_:
  - env-var `PROMNESIA_INDEX_POLICY` (now strick-check its value is one of (`update`|`overwrite`), case-insensitevely)
  - `--update` / `--overwrite` from CLI
  - `--sources` given? (`--update` conveyed)
  - `--overwrite` as fall-back _(changing this would be a breaking change)_.
- Function defaults in the new code are `False`, as suggested in [this comment of #20](https://github.com/karlicoss/promnesia/issues/20#issuecomment-778382929).
- Both `promnesia index` & `promnesia demo` sub-commands have been updated.
- Env-var now checks its value one of (update|overwrite).
- All update/overwrite decision logic moved to __main__.

karlicoss:
Hm, CI fails (for the context, this https://github.com/karlicoss/promnesia/blob/e44a2bda053cc93bb8dc1b84fd1a4461aa5d914f/.github/workflows/main.yml#L40 is the script that github actions are running), but in essence, you can reproduce it localy via `CI=true tox`. 

You can also run specific parts of tox as for example `CI=true tox -e tests`
https://github.com/karlicoss/promnesia/blob/master/tox.ini#L14

But also of course running tests directly with pytest would work too, guess you'll figure this out. But `tox` is the closest approximation to CI.

src/promnesia/__main__.py
```diff
@@ -256,6 +299,24 @@ def main() -> None:
     ep.add_argument('--dry', action='store_true', help="Dry run, won't touch the database, only print the results out")
     # TODO use some way to override or provide config only via cmdline?
     ep.add_argument('--intermediate', required=False, help="Used for development, you don't need it")
+    ep.add_argument('--sources', required=False, action="extend", nargs="+", type=_parse_ordinal_or_name,
+                    help="Subset of source(s) to run (name or 0-indexed position);  use `promnisia --dry` to view sources")
```
karlicoss: nit: typo in "promnesia"

src/promnesia/dump.py
```diff
@@ -30,7 +21,7 @@ def update_policy_active() -> bool:
 
 
 # returns critical warnings
-def visits_to_sqlite(vit: Iterable[Res[DbVisit]]) -> List[Exception]:
+def visits_to_sqlite(vit: Iterable[Res[DbVisit]], overwrite_db: bool) -> List[Exception]:
```
karlicoss: nit: best to use `*, overwrite_db: bool` to enforce named argument (so they aren't confused on the call site)

src/promnesia/__main__.py
```diff
@@ -29,8 +42,16 @@ def iter_all_visits() -> Iterator[Res[DbVisit]]:
 
     hook = cfg.hook
 
-    indexers = cfg.sources
-    for idx in indexers:
+    indexers = list(cfg.sources)
+
+    if sources_subset:
+        indexers = _decide_indexers(indexers, sources_subset)
+    logger.info("sources to extract: %s", ", ".join(i.name for i in indexers))
```
karlicoss: It can be an Exception (e.g. if it's a 'lazy' indexer), in which case `.name` would cause everything to crash. If you run tox -e mypy-core`, it flags it.

IMO best to keep it iterative, without forcing to list, and move the check below `instance(idx, Exception)` check. Worst case the user will have the subset they specified + a bunch of exceptions.

karlicoss:
> PROMNESIA_INDEX_POLICY

To be honest, I think it would make sense to deprecate it (not a major violation of backwards compatibility, doubt many people used it either). But maybe nice to add a `warning.warn` to start with.

src/promnesia/__main__.py
```diff
@@ -309,19 +372,47 @@ def main() -> None:
         p.print_help(sys.stderr)
         sys.exit(1)
 
+    overwrite_policy_var = os.environ.get("PROMNESIA_INDEX_POLICY")
+    if overwrite_policy_var:
+        overwrite_policy_var = overwrite_policy_var.lower()
+        if overwrite_policy_var not in ("update", "overwrite"):
+            print(
```
karlicoss: logger.error?

karlicoss:
Thanks! Left some comments..
Btw think I noticed an issue recently -- I increased the frequency of updates and sometimes noticed that backend is unresponsinve (likely during DB updates). Not directly related to this PR, just something to keep in mind for the future (since `--update` makes sense as default)

src/promnesia/__main__.py
ankostis: Just copied code from [above](https://github.com/karlicoss/promnesia/blob/e44a2bda053cc93bb8dc1b84fd1a4461aa5d914f/src/promnesia/__main__.py#L308),
which didn't have logging.  Additionally, that's an do-it-once "static" error, notprobable to change between runs, so you would detect and fix it the 1st time it rings.
But i can add a log-statement, if you want it?

My personal pattern for errors that should _exit politely with some error-code and a message ("expected errors")_,
is to raise a new exception with those data, and i wrap all CLI code into error-handling+logging just for that exception type - all other exceptions dump stack-trace at the console ("unexpected errors").

I would prefer to submit a separate PR to substitute all such code places, if you are ok?

src/promnesia/__main__.py
ankostis: Fixed & rebase original commit as 92ecac7 [edit: commit id].

src/promnesia/dump.py
ankostis: Done & rebased original commit as 0775abb [edit: commit id].

ankostis:
A kind request, when it's time to merge (not yet), i suggest to rebase this PR - not to squash it,
in order to preserve the commits herein, since they are articulated and isolated as "features", well described accordingly on each commit-message.

src/promnesia/__main__.py
karlicoss: Ah yes good point. Either way, it's a minor thing.

> My personal pattern for errors that should exit politely with some error-code and a message ("expected errors"),
is to raise a new exception with those data, and i wrap all CLI code into error-handling+logging just for that exception type - all other exceptions dump stack-trace at the console ("unexpected errors").

Yeah, this makes sense. I guess also makes it more convenient to test (so you can catch the exception). Not a big issue perhaps, but yeah if you have time for such PR -- sure!

karlicoss:
> i suggest to rebase this PR - not to squash it,

Yes! I have a similar workflow, and happy to do that when people maintain clean history. I think I squashed your first PR since it had lots of 'fix' commits, but the other one I just rebased as is.

src/promnesia/__main__.py
ankostis: Fixed in 0775abb and tests now pass, except those requiring `private` module.

What is this??

ankostis:
### More work...
> To be honest, I think it would make sense to deprecate `PROMNESIA_INDEX_POLICY` (not a major violation of backwards compatibility, doubt many people used it either). But maybe nice to add a `warning.warn` to start with.
> ...
> (since `--update` makes sense as default)

- ok, 2d5c3c4 dropped `--update` option (only `--overwrite` remains), as "update" is now the default behavior, and
- therefore removed that env-var completely, since if anyone had used `PROMNESIA_INDEX_POLICY=update`,
won't notice any difference
(assuming nobody used `PROMNESIA_INDEX_POLICY=no`, and decided that notifying user about this experimental feature being dropped, didn't worth its weight).
- Now `--sources` option is uncoupled from `--overwrite` (much simpler CLI to grasp).
- OP has been updated to reflect the final functionality of this PR.

 ###  Still owing ...
- Some test-cases on the new functionality,
  if only i could make them all run (see my question about `private` module, above)??
- Also please consider #213 (black-en sources) and adding `CHANGES.md`.

src/promnesia/__main__.py
karlicoss: Right, checked the 'import private' things -- these are automation scripts in [demos.py](https://github.com/karlicoss/promnesia/blob/master/tests/demos.py ) I used to record demos in [readme](https://github.com/karlicoss/promnesia#readme). They are also serve as kind of end to end tests, however . The private thing was because they imported some cookies which I didn't want to commit in the main repository.
Either way they are not running on github because of this: https://github.com/karlicoss/promnesia/blob/e44a2bda053cc93bb8dc1b84fd1a4461aa5d914f/tox.ini#L23

The `not with_browser` thing excludes tests marked with `with_browser` (so all tests in `demos.py`).
What was the command you used to run? Either way, sorry it's confusing, I guess a better option would be to make the default to run the same what CI is running, and only if user want, explicitly make it run these GUI tests.

There is also another test guilty of this: https://github.com/karlicoss/promnesia/blob/e44a2bda053cc93bb8dc1b84fd1a4461aa5d914f/tests/indexer_test.py#L158-L159 -- it doesn't really run on CI, because it has `CI` env variable set, but it might not be the case for people.

In the meantime, the most reliable command to run is `CI=true tox -e tests`, and you can also restrict them to specific test like this: `CI=true tox -e tests -- -k test_demo` (it's the one failing). I'll try to clean up the CI config in the meantime so it's easier for people to contribute.

karlicoss: Similar for mypy: `CI=true tox -e mypy-core` -- also has a couple of failures

karlicoss:
For #213 -- will respond there
For changelog -- yeah I agree. Let's discuss at #215

src/promnesia/__main__.py
karlicoss: Ok, I cleaned it up a bit https://github.com/karlicoss/promnesia/pull/216
So if you rebase, hopefully you'll be able to simply run the tests with
- `tox` to run all of them (tests + mypy)
- `tox -e tests` to just run unit tests
- `pytest tests` to just directly via pytest

ankostis: Great!

Hopefully i fixed the bug in `do-demo()` and mypy issues, but
`tests/cli.py::test_demo` still fails with this:

```
do_demo( index_as=getattr(args, 'as'), params=args.params, port=args.port, config_file=args.config, name=args.name, sources_subset=args.sources)
*** FileNotFoundError: [Errno 2] No such file or directory: '~/.config/promnesia-test/config.py'
Traceback (most recent call last):
  File "./promnesia.git/src/promnesia/__main__.py", line 145, in do_demo
    config.load_from(config_file)
  File "./promnesia.git/src/promnesia/config.py", line 102, in load_from
    instance = import_config(config_file)
  File "./promnesia.git/src/promnesia/config.py", line 118, in import_config
    spec.loader.exec_module(mod) # type: ignore
  File "<frozen importlib._bootstrap_external>", line 786, in exec_module
  File "<frozen importlib._bootstrap_external>", line 922, in get_code
  File "<frozen importlib._bootstrap_external>", line 979, in get_data
```

Any idea how ` ~/.config/promnesia-test/config.py` file came to be?

ankostis: Also i would like some recommendations on how to structure the test-cases for the new CLI features (the 2 new options), like, which files to create, and what to patch, etc.

And if i get it right, there were no new test-cases for the  "update db' functionality you did in #20, correct @karlicoss.  would t be possible for you to do it?

src/promnesia/__main__.py
```diff
@@ -61,17 +77,23 @@ def it() -> Iterable[Res[DbVisit]]:
...
             errors.append(e)
     return errors
 
 
-def do_index(config_file: Path, dry: bool=False) -> None:
+def do_index(
+        config_file: Path,
+        dry: bool=False,
+        sources_subset: Iterable[Union[str, int]]=(),
+        overwrite: bool=None,
```
karlicoss: seems that 'overwrite' is redundant?

src/promnesia/__main__.py
```diff
@@ -269,13 +328,13 @@ def main() -> None:
     ap.add_argument('--name', type=str, default='demo'               , help='Set custom source name')
     add_port_arg(ap)
     ap.add_argument('--no-serve', action='store_const', const=None, dest='port', help='Pass to only index without running server')
-    ap.add_argument('--config', type=Path, required=False            , help='Config to run against. If omitted, will use empty base config')
```
karlicoss, line 272: Hmm, I think the `test_demo` failure was caused by this, because here the config argument is optional (which means that it would use the temporary directory, see [here](https://github.com/karlicoss/promnesia/blob/65e77cd67740ef391cba73bc9eac200d9a5b6e63/src/promnesia/__main__.py#L110-L121) ).
Whereas in `index` there is `default_config_path()`, this is because under tests we want to avoid running against the actual user config, so it uses a different directory (`promnesia-test`).

I kinda see your point about reusing the arguments though -- in principle indexing is part of `demo` mode, so this makes sense.. Maybe a 'cheap' way to fix it would be to pass `default_config` to `add_index_args`? So index would use `default_config_path()`, and adhoc will pass `None`, and later we can think what would be a 'proper' way to unify them?

karlicoss (changes requested):

src/promnesia/__main__.py
karlicoss: > Any idea how ~/.config/promnesia-test/config.py file came to be?

Yeah, commented about the test https://github.com/karlicoss/promnesia/pull/211#discussion_r588912230

> Also i would like some recommendations on how to structure the test-cases for the new CLI features 

Sure! For tests 'structure', it's currently:

- `tests/indexer_test`  -- usually tests sources and how they extract visits
- `tests/integration_test` -- usually tests indexing, and does actual database checks
- `tests/server_test` -- usually tests database responses
- `tests/cli` -- tests cli interation

I say 'usually' because it's a little messy and the separation is not 100% clear at times. For example, most tests in `server_test` also usually run indexing and extraction, they aren't very 'unit' tests. Not against it, but just with restricted resources full-scale tests give better coverage, at the expense of a bit more confusion when they fail.

So, I only recently started CLI tests in `cli.py` -- and it's actually the one failing now ;) In general would be nice to have more tests there. I think the most important thing is so things don't crash there -- the actual code in `__main__` is fairly covered by other tests.

So maybe adding a couple of tests which do `check_call(promnesia_bin(command_string.split()))` or something like that wouldn't hurt. Probably the most important candidates are commands already mentioned in README and other docs -- would be the worst to break them. 

If any command needs data source, can use `ox_hugo_data` which is already in this file, or `hypothesis_data` which is used in `integration_test`. Note that you might need to run `git submodule update --init --recursive` -- the test data is in submodules (to avoid bloating the main repository). Let me know if you have any issues with this!

> And if i get it right, there were no new test-cases for the "update db' functionality you did 

For tests, at first I thought that I didn't, but it seems there is one! 
https://github.com/karlicoss/promnesia/blob/master/tests/server_test.py#L178
However it's not actually testing anything meaningful, just checks nothing crashes. 

In fact now it would be even less meaningful, because of (unfortunately) hardcoded env variable.
https://github.com/karlicoss/promnesia/blob/master/tests/integration_test.py#L14-L24

I don't mind adding it after we merge though -- I could share a PR with you so you could see what I change and ask questions to make yourself more familiar. Sounds good?

src/promnesia/__main__.py
ankostis: Nice catch, thanks, fixed.

src/promnesia/__main__.py
ankostis: Completely missed the different values...so i'm either too old to program, or it's the 5 month curfew here in Greece that has taxed heavily on my mental health, or both :-/
Anyway, fixed it, and can't think of another 'proper' way, this seems to be just right, added some documentation that seem appropriate.

karlicoss: @ankostis no worries, took me a little to figure out as well even though I wrote the code, so don't think it's the age :) 
(also sorry to hear about the curfew! we don't have that in London, just lockdown and it already sucks, can't imagine when it's also with curfew rules)

src/promnesia/__main__.py
```diff
@@ -245,15 +276,46 @@ def cli_doctor_server(args: argparse.Namespace) -> None:
...
+        s = int(s)  # type: ignore
+    except ValueError:
+        pass
+    return s
+
+
 def main() -> None:
     # TODO longer, literate description?
 
+    def add_index_args(parser: argparse.ArgumentParser, default_config_path: PathIsh) -> None:
```
karlicoss: default_config_path should be `Optional[PathIsh]` since can be None (mypy will complain)

src/promnesia/__main__.py
ankostis: Noticed that it fails in python<3.8 due to argparse `action='extend'`...i can downgrade it or hope that 3.7 is dying a slow death?

src/promnesia/__main__.py
```diff
@@ -245,15 +276,46 @@ def cli_doctor_server(args: argparse.Namespace) -> None:
...
+        """
+        :param default_config_path:
+            if not given, all :func:`demo_sources()` are run
+        """
+        parser.add_argument('--config', type=Path, default=default_config_path, help='Config path')
+        parser.add_argument('--dry', action='store_true', help="Dry run, won't touch the database, only print the results out")
+        parser.add_argument(
+            '--sources',
+            required=False,
+            action="extend",
```
karlicoss, line 301: hmm, so CircleCI test failed (sorry for some reason it requires github login -- will figure out if it's possible to make it public, or even migrate to github actions)
```
Traceback (most recent call last):
  File "/opt/circleci/.pyenv/versions/3.7.0/lib/python3.7/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/opt/circleci/.pyenv/versions/3.7.0/lib/python3.7/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/home/circleci/project/src/promnesia/__main__.py", line 413, in <module>
    main()
  File "/home/circleci/project/src/promnesia/__main__.py", line 318, in main
    add_index_args(ep, default_config_path())
  File "/home/circleci/project/src/promnesia/__main__.py", line 304, in add_index_args
    help="Source names (or their 0-indexed position) to index.",
  File "/opt/circleci/.pyenv/versions/3.7.0/lib/python3.7/argparse.py", line 1352, in add_argument
    raise ValueError('unknown action "%s"' % (action_class,))
ValueError: unknown action "extend"
Traceback (most recent call last):
  File "tests/install_and_run", line 80, in <module>
    main()
  File "tests/install_and_run", line 76, in main
    run(Path(tdir))
  File "tests/install_and_run", line 36, in run
    check_call([promnesia, 'index', '--config', cfg_file])
  File "/opt/circleci/.pyenv/versions/3.7.0/lib/python3.7/subprocess.py", line 328, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '[PosixPath('/home/circleci/project/scripts/promnesia'), 'index', '--config', PosixPath('/tmp/tmp7nmhmvz6/config.py')]' returned non-zero exit status 1.
```
I think the reason is that there is no `extend` action in Python prior to `3.8` -- and promnesia runs on 3.6, 3.7 and 3.8 https://github.com/karlicoss/promnesia/blob/master/.github/workflows/main.yml#L20

Not sure what's the difference with `append` action? But if you think extend is more useful, quick googling suggests it's easy to supprort https://stackoverflow.com/a/41153081/706389
perhaps `compat.py` would be a good place for it.

src/promnesia/__main__.py
karlicoss: Ah, you were 10 seconds ahead of me, I left the comment :) 
3.6 maybe almost dead? But I think 3.7 would be a little premature, many linux/osx installs have it. (even circleci, apparently!)

src/promnesia/__main__.py
ankostis: Right, i made it into an optional arg, which conveys better its usage.

src/promnesia/__main__.py
ankostis: Converting to `append`, you lose this syntax: `--sources viber telegram`,
and forced to use `--source viber --source telegram`
(notice that then the options cannot be in plural).

Will add SO workaround in `compat.py`.

ankostis: Just pushed and waiting for circleCI's opinion.

Locally, my tox fails with ` ModuleNotFoundError: No module named 'dal_helper'`.
What should id do?

ankostis: OK, done!
(in a separate commit, to isolate the compatibility issue).

karlicoss: > ModuleNotFoundError: No module named 'dal_helper'
I think it would also besolved by `git submodule update --init --recursive`

Hmm. So all tests passing locally for me! But now github fails in weird ways.. I'll try to investigate why. Feels like some intermittent github issue though.

src/promnesia/__main__.py
ankostis: Will do my study, and come back,
or you may take over?

karlicoss: I think I found it! Was missing the directory, and sql seemed very obscure about it
https://github.com/karlicoss/promnesia/commit/f47c03e63582adb70f18b948d99f3fc92f4dcf54
I could merge and then you rebase, or might be just easier if you cherry pick the commit? 

Curious why it didn't happen before, but probably a different test order or something like that? (sometimes it depends on who knows what). Either way probably not worth investigating too much -- in my experience empty dir never hurts anyway, and eases a lot of pain like that.

If you're curious how I figured this out -- there is a thing in the github actions pipeline one can uncomment https://github.com/karlicoss/promnesia/blob/master/.github/workflows/main.yml#L36-L37
after that it's possible to SSH onto the machine tests are running on and reproduce. When I reran the test the second time and they worked I started suspecting something :)

ankostis: Can cherry-pick ~~(have not yet undestand what is it about)~~.

karlicoss: All looks good! Happy to merge?

ankostis: Not yet, need to retrofit the old test-case with the env-var.

ankostis: OK, done, please merge.

(i noticed that GitHub run just 1 check o the PR, did you change something in the actions?

ankostis: ignore my last comment about missing GH-actions.

karlicoss: No, you were right! I have to manually push it at the moment to trigger (to a different branch, as long as it has the same commit hash, it shows the pipelines).
I experimented and think I figured out how to set it up properly so it works on forked repos/PRs as well, I'll do it later & update the instructions.

karlicoss:
Hooray, thanks!

Sorry it's taken a while, I guess todos for me which would make it quicker next time:
- [x] make github actions run on pull requests & forks
- [x] add a section to development.org about running tests (tox/mypy/ `-k` argument) & basic explanation of what's in each of them (maybe even docstring is good enough) + mention submodules update command explicitly
- [x] (maybe) get rid of circleci in favor of github actions, just less external dependencies and makes it easier to run on forks as well

I'll look at it after I get some sleep (crazy schedule)

Other todos from the discussions:
- [x] add meaningful test for update functionality (I can do it and share the PR?)
      in particular, there might be a minor race condition (database unresponsive during the update), would be nice to cover it with regression test
- [ ] add some basic cli tests
- [ ] think about unifying cli commands in the future, since they all kind of do similar things

Not sure if forgot anything?

karlicoss:
Done the CI updates & developer info! https://github.com/karlicoss/promnesia/pull/217 keeping circle though, since it's still useful because of systemd test.

Also if you go to “Actions” tab on your promnesia fork and click “I understand my workflows, go ahead and enable them”, the pipelines should start running for you even before you PR, so hopefully will make it all quicker next time.

src/promnesia/__main__.py
ankostis: > IMO best to keep it iterative, without forcing to list, and move the check below instance(idx, Exception) check. 
> Worst case the user will have the subset they specified + a bunch of exceptions.

A note from posterity: i should have kept the case where `--sources` is given as a `list`, as originally implemented by `_decide_indexers()`) - not really gaining much memory by being iterative, sources are not gonna be innumerable.  And would avoid seeing a bunch of irrelevant lines everytime i update a single source.

See also discussion [about iterables in zulip](https://memex.zulipchat.com/#narrow/stream/279601-hpi/topic/Iterator.20vs.20Iterable/near/231798316).

PR #222, opened 2021-03-31
enh: harvest URLs from Hypothes.is comments & higleted text

ankostis:
... plus a minor simplification-refactoring.

### DIFF
--- a/src/promnesia/sources/hypothesis.py
+++ b/src/promnesia/sources/hypothesis.py
@@ -4,16 +4,20 @@ Uses HPI [[https://github.com/karlicoss/HPI/blob/master/doc/MODULES.org#myhypoth
 from ..common import Loc, Results, Visit, extract_urls
 
 
-def _harvest_text(text, visit_base) -> Results:
+def _harvest_text(text, part_name, visit_base) -> Results:
     if text and text.strip():
         urls = extract_urls(text)
         for url in urls:
-            yield visit_base._replace(url=url)
+            yield visit_base._replace(
+                url=url,
+                locator=visit_base.locator._replace(title=f"hypothesis-{part_name}"),
+            )
 
 
 def index() -> Results:
     from . import hpi
     import my.hypothesis as hyp
+
     for h in hyp.get_highlights():
         if isinstance(h, Exception):
             yield h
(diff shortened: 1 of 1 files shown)


karlicoss:
Good idea, thanks!

src/promnesia/sources/hypothesis.py
```diff
@@ -1,12 +1,23 @@
-'''
+"""
 Uses HPI [[https://github.com/karlicoss/HPI/blob/master/doc/MODULES.org#myhypothesis][hypothesis]] module
-'''
-from ..common import Visit, Results, logger, Loc
+"""
+from ..common import Loc, Results, Visit, extract_urls
+
+
+def _harvest_text(text, part_name, visit_base) -> Results:
```
karlicoss, line 7: small nits: 
- I usually try to keep these helpers below 'index' (so you see 'index' the first thing) when you open the module,which is the most important
- if there are multiple arguments of same type (like text/part_name), I usually force them to be keyword (i.e. `def _harvest_text(*, text, part_name, visit_base)`), otherwise it's easy to confuse them and not notice (and even `mypy` won't help!)
but not big issues, just sharing!

ankostis:
> I usually try to keep these helpers below 'index' 

I follow call-dependency order, so that contradicted with the UX-order you follow - will keep that in mind next time, and put exclusively `index()` at the top, ok?
 
> if there are multiple arguments of same type (like text/part_name), I usually force them to be keyword 

I would rewrite this PR to inline the helper-function, and eliminate both nits, but now its been merged `¯\_(ツ)_/¯`

karlicoss:
> ill keep that in mind next time

Thanks! No problem anyway, it makes sense too! Like I said more of a nit.

> I would rewrite this PR to inline the helper-function, and eliminate both nits, but now its been merged 

IMO it's better to merge sooner than later for faster iteration if it doesn't break things :) But I can ask first next time if you prefer?

src/promnesia/sources/hypothesis.py
ankostis: Ok, done in fba2044 from #224, by inlining the helper-function (as [suggested here)](#issuecomment-811776147).

PR #223, opened 2021-03-31
Harvest Telegram http only URLs

ankostis:
Use an SQLite criteria to fetch chats containing only `"%http%",
which excludes plain IPs, strings like `"python.py"` to be harvested as visits..

Actually the code was commented-out, as not offering any speed-improvement,
but in my case, harvesting runs ~x4 times faster.


karlicoss:
Yeah indeed seems to take more time, about 3x for me..
But also seems to result in different number of visits (about 15% less!). I checked the diff and I remember now why I commented it out.
Indeed -- some of it ends up as garbage like `BlockingObservable.java:416`. But also sometimes people send a plain domain name (e.g. `www.website.com`) and it doesn't get recognized by telegram as media with URL etc for some reason... even though it's highlighted as blue in the app. So I guess we shouldn't filter (but also perhaps worth updating the comment to reflect this info).

Do you experience any particular issues with it? The extra stuff like `python.py` shouldn't hurt, except for bloating database a tiny bit. And if it's slow perhaps we can think about parallelizing it -- it's been long coming anyway (the slow part is not the query, but URL extraction which would be fairly easy to parallelize for telegram at least).

ankostis:
- Doesn't have an effect a boated db on page-loading times?
- Other modules already do this `http` filtering - why break this convention only in this source?

karlicoss:
- no, not on page load times, since the data is requested after the page is loaded (to be more specific, its URL), and in parallel. And again, it's not a massive blowup (in return for some extra useful data), plus the most time is spent showing visits (i.e. rendering), the DB queries are cheap and likely dominated by IO latency...
  And if this is noticeable, I'd much rather prefer filtering out false visits better (e.g. by excluding things that looks like source code files). It's already possible via `HOOK` config setting, for example -- just could also work on doing it in a generic way so it works for everyone
- `extract_url` function doesn't, it can handle links without the protocol! And if it does happen in some modules, most likely it's by accident.

ankostis:
Yes, i broadly referred to "page-loading" time, while decorating URLs happens in an async task (correct?), so even though the page has been loaded, you still see JS running if there are many URL, no?  Anyhow, the db size and db-queries are just a resource-waste issue (storage, CPU), not latency.  

The same applies for parallelizing the index operation - the CPU & storage would still be there, just faster to complete;  but do we really want to occupy all cpu-cores for a background task?

In my efficient-hungry mind (thin of it as CO2 footprint), eliminating bloat is always good.
Additionally, for me it's more pronounced since i have many internal code-samples with IP lists, so i get  more that x2 vistis, and x4+ index-time, so i'd love to add a **global configuration option** (like `HOOK`) to control whether lose-parsing of URL-likes is enabled or not.
What do you think?

ankostis:
Just to clarify that i'm proposing another global config, like `HOOK` - the later is not satisfactory, because it doubles the cpu-effort, once to harvest all URLs (incl. false- visits) and later to filter those out.

karlicoss:
> while decorating URLs happens in an async task (correct?),

Technically decorating itself is on the UI thread (because DOM modifications run on the same thread). But requests/other processing is in the background, yeah. 

> so even though the page has been loaded, you still see JS running if there are many URL, no?

There are two cases when 'false positive' URLs in database can slow things down
- if these fake URLs are present on the page
  However this would require links like `http://python,py` which of course doesn't happen in practice
- if there is a really stupid amount of these 'fake' urls in the database.
  I'd struggle to even estimate how much would you need to have impact... like gigabytes of false positives? 

> In my efficient-hungry mind (thin of it as CO2 footprint), eliminating bloat is always good.
 
Agree! But using Promnesia is a choice in the first place, so it means already accepting some bloat in our life :) Of course, hopefully it will have net positive because it will empower us to be better at retaining knowledge to fight for important things, including CO2 footprint, which I also care about! 
Basically here the problem is that there is some tradeoff between efficiency and power of the tool (ability to capture more useful URLs at cost of having falst positives & bloat), so I don't think that either of us is 'more correct' anyway -- it's a somewhat arbitrary choice after all. Maybe making it configurable would be a good compromise? So you could run the indexer with extra flag which omits these visits? 

Perhaps one other thing is that imho it's best to optimize bloat after doing some profiling (especially when it's not super obvious like here and comes at cost of power of the tool). Otherwise we're just handwaving because we don't know how impactful would each change be :)

> do we really want to occupy all cpu-cores for a background task?

Not all cores, it would be configurable, so would be up to you. I'd definitely love it to work faster for me! 

> Just to clarify that i'm proposing another global config, like HOOK 

Yeah, I see your point, it makes sense. In fact there is such a thing, `FILTER` in config, although I've been thinking to sunset it in favor of `HOOK`. 

Here's where filtering happens: https://github.com/karlicoss/promnesia/blob/ca61d96b3a42d620bda812c5d290ee048cb08ccc/src/promnesia/extract.py#L86-L87

Here's a test for this
https://github.com/karlicoss/promnesia/blob/ca61d96b3a42d620bda812c5d290ee048cb08ccc/tests/indexer_test.py#L231-L254

It accepts regexes or predicates, so hopefully would be easy for you to filter out garbage in the meantime?
Ultimately I agree it would make sense for this filtering to happen somewhere on `extract_urls` level? So we don't even construct `Visit` objects. But also want to keep the config minimal at the same time... I guess I'll keep filter for now but let's also think how to make it the most flexible (without rushing to implement for now).

ankostis:
I added a configuration argument `http_only`, to preserve the capability AND preserve old behavior by default.
Should that become keyword-only?

ankostis:
I also fixed the sqlite db to open in a managed-context, so as to close the file after the source has run.

ankostis:
Had also to retrofit _viber source_ to behave as the rest sources,
which i had originally constructed to harvest http-only links,
so i added also there the new `http_only` config argument.

karlicoss:
thanks, looks good!

PR #224, opened 2021-03-31
Seprate Viber tags with spaces

ankostis:
...like hypothesis-source does.

Also use [JSON1 SQLite](https://www.sqlite.org/json1.html) extension to extract items from a JSON column.

### DIFF
--- a/src/promnesia/sources/hypothesis.py
+++ b/src/promnesia/sources/hypothesis.py
@@ -1,17 +1,7 @@
 """
 Uses HPI [[https://github.com/karlicoss/HPI/blob/master/doc/MODULES.org#myhypothesis][hypothesis]] module
 """
-from ..common import Loc, Results, Visit, extract_urls
-
-
-def _harvest_text(text, part_name, visit_base) -> Results:
-    if text and text.strip():
-        urls = extract_urls(text)
-        for url in urls:
-            yield visit_base._replace(
-                url=url,
-                locator=visit_base.locator._replace(title=f"hypothesis-{part_name}"),
-            )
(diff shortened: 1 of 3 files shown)


src/promnesia/sources/hypothesis.py
```diff
@@ -1,12 +1,13 @@
...
 '''
-from ..common import Visit, Results, logger, Loc
+from ..common import Loc, Results, Visit, join_tags, logger
 
 
 def index() -> Results:
-    from . import hpi
     import my.hypothesis as hyp
+
+    from . import hpi
```
karlicoss: nope, `from . import hpi` has to be first, otherwise `import my.hypothesis` fails with obscure error if HPI isn't installed

karlicoss (changes requested):
Looks good, just one comment!
`macos-latest, 3.6` pipeline failed, but looks like a random github glitch, will probably go away on next change

src/promnesia/sources/hypothesis.py
ankostis: Thanks!.  Fixed in 429d882.

ankostis: But shouldn't these import-error manifest themselves?

karlicoss: Yep, in that sense it's redundant -- it just gives a bit friendlier error message for people less familiar with python

karlicoss:
looks ok, but think it conflicts with the previous PR merged before -- so needs a rebase/conflict resolution

ankostis:
Resolved conflict and did some suggested nit picks in #222.

karlicoss:
thanks!

PR #227, opened 2021-04-14
firefox updates

karlicoss:


purarue:
Uh, just curious

was there any reason to include this in the shared `_row2visit_firefox` function and not create separate handlers for the datetime part of Firefox and FirefoxPhone?

Just less ability to share it as a classmethod?

karlicoss:
So interesting enough, Fenix (beta Firefox) actually has the same schema as desktop Firefox (so already was matching `class Firefox`, not `class FirefoxPhone`).
But still seems that there might be subtle schema differences like this one :shrug: Figured it wasn't worth duplicating most of `class Firefox` to `class FirefoxPhoneNew` or something like that.

whoops, just noticed accidental print https://github.com/karlicoss/promnesia/blob/0e1e9a1ccd1f07b2a64336c18c7f41ca24fcbcd4/src/promnesia/sources/browser.py#L212
wonder if this is how you noticed it xD

purarue:
Ah, no, I've still been trying to figure out how to backup my firefoxphone db using termux, am very new to that

as im currently using a mix of [ffexport](https://github.com/seanbreckenridge/ffexport) and some ['private' promnesia stuff](https://github.com/seanbreckenridge/HPI/blob/master/my/chrome.py#L28) to merge my browsing history in HPI, is getting a bit annoying to manage...

Can't really just use promnesia because I want access to the description/title/image that firefox exposes; but understandably that wouldn't be needed in promnesia

So, spent a few hours [restructuring ffexport](https://github.com/seanbreckenridge/ffexport/tree/v0.2/browserexport/browsers) to a namespace package to allow it to be more extendible, so it supports firefox derivates etc, was reading over the code here and noticed the recent changes

purarue:
I guess one major difference is that the `Extr` class Im defining also [describes how to backup the databases](https://github.com/seanbreckenridge/ffexport/blob/v0.2/browserexport/browsers/firefox.py#L36-L46)

But otherwise that project and promnesia are pretty similar, some code duplication as far as detecting/reading from databases goes

Is what I was using before promnesia becuase I still wanted to backup my history, but promnesia seemed a bit daunting whilst I was still building HPI out

Plan to create a similarieis/difference thing in the readme

karlicoss:
Nice! I mean if it's flexible enough then `ffexport` can just be used in promnesia, right? Would be cool to move this outside.

purarue:
Once its a bit more stable than it is currently, I wouldn't be opposed to trying to replace `browser.py` and the [browser_history.py](https://github.com/karlicoss/promnesia/blob/master/scripts/browser_history.py) script

ankostis:
Indeed now `browser_history.py` is deprecated, but i cannot find instructions how to accomplish the same promnesia-update task with [`browser.py`](https://github.com/seanbreckenridge/browserexport ?

PR #228, opened 2021-04-14
fix spelling mistakes in comments

purarue:


karlicoss:
thanks... I should finally set something up in my emacs :sweat_smile:

PR #229, opened 2021-04-15
support specifying config file through envvar

### LINKED ISSUE #226: support specifying config dir through an environment variable
Similar to HPI, would be nice if this was possible.

https://github.com/karlicoss/HPI/blob/c1b70cd90e894835d18d9cac04cc1d12724bf3c7/my/core/preinit.py#L8

Else, on mac this puts my config somewhere in /Application Support/ and syncthing cant sync my promnesia config across my computers...

Currently this just uses appdirs without letting the user specify otherwise...

https://github.com/karlicoss/promnesia/blob/137c3d2e685c08a4ead3c2602d7f6cdaf877ed1f/src/promnesia/common.py#L518-L519

Sort of wished appdirs just used XDG_CONFIG_HOME on mac as well...

do you think `PROMNESIA_CONFIG` works as the name of the environment variable?

Am willing to make a PR at some point soon

purarue:
resolves #226

### DIFF
--- a/README.org
+++ b/README.org
@@ -234,7 +234,7 @@ Also check out [[file:doc/TROUBLESHOOTING.org][troubleshooting guide]] or [[http
 
 - create the config: =promnesia config create=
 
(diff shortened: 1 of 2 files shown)


karlicoss:
thanks!

Issue #233, opened 2021-04-20
render markdown in github/reddit context


purarue:
Following up on from the discussion from #232 

~~Tried an initial attempt at trying to render the markdown for `my.discord` context using `mistletoe.Document.render` in the backend, but I don't think it wouldn't scale well - especially if you have a couple thousand github/reddit comments - would estimate it would take close to an 30 minutes just to index...~~

Edit: wasn't `mistletoe` causing the lag, see comment below

Perhaps a flag could be added to `github` like

```
def index(render_markdown: bool = False):
 ....
```

to optionally enable it?

---

The other option is to just tag it as markdown in the `promnesia index` phase with another hint like !markdown and check for that while rendering the tag in the extension.

On any singular page, the amount of reddit/github comments found by promnesia is relatively low (so typically wouldn't cause lots of lag?) compared to something like browser history, so this may be the better strategy -- would likely need a library to handle that in JS. Should probably have a option to enable/disable that as well

Not a high priority issue, may try to see what I can figure out once I'm a bit more familiar with building/developing on the browser extension here

purarue:
Not sure if the backend strategy is viable, but ended up writing this helper to extend the `promnesia.sources.markdown` module:

```diff
diff --git a/src/promnesia/sources/markdown.py b/src/promnesia/sources/markdown.py
index af1d338..2b9bde2 100644
--- a/src/promnesia/sources/markdown.py
+++ b/src/promnesia/sources/markdown.py
@@ -99,3 +99,39 @@ def extract_from_file(fname: PathIsh) -> Iterator[Extraction]:
                 locator=Loc.file(fname), # TODO line number
                 context=r.context,
             )
+
+
+class TextParser(Parser):
+    '''
+    Used to extract links/render markdown from text, e.g. reddit/github comments
+    Instead of chunking blocks like for files, this returns the entire
+    message rendered as the context
+    '''
+    def __init__(self, text: str):
+        self.doc = mistletoe.Document(text)
+
+
+    def _doc_ashtml(self):
+        '''
+        cached html representation of the entire html message/document
+        '''
+        if not hasattr(self, '_html'):
+            self._html = HTML_MARKER + _ashtml(self.doc)
+        return self._html
+
+
+    def _extract(self, cur, last_block = None) -> Iterator[Parsed]:
+        if not isinstance(cur, (AutoLink, Link)):
+            return
+
+        yield Parsed(url=cur.target, context=self._doc_ashtml())
+
+
[...]

PR #234, opened 2021-04-20
sources.markdown: add extract_from_text helper

purarue:
related #233

even if the strategy of converting markdown into HTML on the backend
isn't merged, can see this being useful for other people
writing their own sources for promnesia (e.g. me, for my discord module)

### DIFF
--- a/src/promnesia/sources/markdown.py
+++ b/src/promnesia/sources/markdown.py
@@ -1,4 +1,3 @@
-from datetime import datetime
 from pathlib import Path
 from typing import Iterator, NamedTuple, Optional
 
@@ -6,7 +5,6 @@ from ..common import get_logger, Extraction, Url, PathIsh, Res, Visit, Loc, file
 
 
 import mistletoe # type: ignore
-from mistletoe import Document # type: ignore
 from mistletoe.span_token import AutoLink, Link # type: ignore
 import mistletoe.block_token as BT # type: ignore
 from mistletoe.html_renderer import HTMLRenderer # type: ignore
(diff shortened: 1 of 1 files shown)


purarue:
Have integrated this with the `sources.github` as well, but not sure if you wanted to merge that here, or if you wanted to do something to render it in JS instead

As an example:

[This random comment](https://github.com/karlicoss/HPI/issues/46#issuecomment-683500101) from HPI renders like this in the sidebar when converted from markdown to HTML:



Also worth mentioning that the `<pre>` blocks break the horizontal scrolling

image)

purarue:
Wasnt sure where to add tests for this, so just to show it works;

```python
In [1]: from promnesia.sources.markdown import extract_from_text

In [2]: import itertools

In [4]: import my.github.all
In [5]: for x in itertools.islice(my.github.all.events(),3):
   ...:     if x.body is not None:
   ...:         for url in extract_from_text(x.body):
   ...:             print(url)
   ...:
Parsed(url='https://github.com/seanbreckenridge/dotfiles/commit/4c57fd97cbb2605e63d0cf5d2af37039fe6e6d35', context='!html <p>Note: This is used for <a href="https://github.com/seanbreckenridge/dotfiles/commit/4c57fd97cbb2605e63d0cf5d2af37039fe6e6d35">gitopen</a></p>\n')
```

karlicoss:
Nice, thanks! 

> Also worth mentioning that the pre blocks break the horizontal scrolling

Ah yeah -- I guess what I meant 'something like pre' -- it should be tweakable with an ad-hoc style maybe like `nowrap` or something like that.

PR #235, opened 2021-04-21
sources.github: optional flag to render markdown

purarue:
allows the user to pass the 'render_markdown'
flag to convert the text in the github body (typically
markdown) to HTML, so it renders nicely in the sidebar

disabled by default, as it might break is inline highlights

---

was a bit unsure how to add this to reddit without adding a
bunch of kwargs like `render_markdown: bool = False` to 
[all the functions here](https://github.com/karlicoss/promnesia/blob/master/src/promnesia/sources/reddit.py#L51-L100)..
Also the `TextParser` import should probably happen inside the `index` function...
but then it needs to be passed to `_from_common` somehow.

Perhaps could just use a global var which is only set in `index` and then
used in `_from_common` if `render_markdown` is True

not sure what the best way to solve this, will think about it...

### DIFF
--- a/src/promnesia/sources/github.py
+++ b/src/promnesia/sources/github.py
@@ -2,6 +2,8 @@
 Uses [[https://github.com/karlicoss/HPI][HPI]] github module
 '''
 
+# Note: requires the 'mistletoe' module if you enable render_markdown
+
 from typing import Optional, Set
 
 from ..common import Results, Visit, Loc, iter_urls, logger
--- a/src/promnesia/sources/github.py
+++ b/src/promnesia/sources/github.py
@@ -2,12 +2,22 @@
 Uses [[https://github.com/karlicoss/HPI][HPI]] github module
 '''
 
-from ..common import Results, Visit, Loc, iter_urls
+from typing import Optional, Set
+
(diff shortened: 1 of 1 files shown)


purarue:
Hmm

[Github Status](https://www.githubstatus.com/history) doesnt indicate an outage with Github Actions, the [run failed](https://github.com/karlicoss/promnesia/actions/runs/769653051) with:

`GitHub Actions has encountered an internal error when running your job.`

May be a temporary thing, not sure [if you can re-run the action](https://docs.github.com/en/actions/managing-workflow-runs/re-running-a-workflow)

Let me know if I should push something else to this branch/can squash it later

purarue:
Ah, nevermind -- they posted an incident report: https://www.githubstatus.com/incidents/cj7gzzj30411

karlicoss:
Thanks, looks good!

> was a bit unsure how to add this to reddit without adding a bunch of kwargs like render_markdown: bool = False to

hmm yeah -- I guess kind of asks for maybe a global setting, or per-module settings. Or maybe put all `_from` things in a helper class and keep the setting as a field?

PR #236, opened 2021-04-22
sources.reddit: add flag to render markdown body

purarue:
Related:
https://github.com/karlicoss/promnesia/issues/233
https://github.com/karlicoss/promnesia/pull/235

### DIFF
--- a/src/promnesia/sources/reddit.py
+++ b/src/promnesia/sources/reddit.py
@@ -3,9 +3,48 @@ Uses HPI [[https://github.com/karlicoss/HPI/blob/master/doc/MODULES.org#myreddit
 '''
 
 from itertools import chain
-from typing import Set
+from typing import Set, Optional
 
 from ..common import Visit, Loc, extract_urls, Results, logger
+
+
+def index(*, render_markdown: bool = False, renderer: Optional['RedditRenderer'] = None) -> Results:
+    from . import hpi
+    from my.reddit import submissions, comments, saved, upvoted
+
+    if renderer is not None:
+        assert callable(renderer), f"{renderer} is not a callable (should be a subclass of RedditRenderer)"
+        r = renderer(render_markdown=render_markdown)
+    else:
+        r = RedditRenderer(render_markdown=render_markdown)
+
+    logger.info('processing saves')
+    for s in saved():
+        try:
+            yield from r._from_save(s)
+        except Exception as e:
+            yield e
+
+    logger.info('processing comments')
+    for c in comments():
+        try:
+            yield from r._from_comment(c)
+        except Exception as e:
+            yield e
+
+    logger.info('processing submissions')
+    for sub in submissions():
+        try:
+            yield from r._from_submission(sub)
+        except Exception as e:
+            yield e
+
+    logger.info('processing upvotes')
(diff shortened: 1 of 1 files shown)


src/promnesia/sources/reddit.py
```diff
@@ -3,40 +3,153 @@
...
+            emitted.add(url)
+
+        # extract from markdown links like [link text](https://...)
+        # incase URLExtract missed any
+        #
+        # this should try to do this, even if the user didn't enable
+        # the render_markdown flag, as it may catch extra links that URLExtract didnt
+        # would still require mistletoe to be installed, but
+        # the user may already have it installed for the auto/markdown modules
+        if hasattr(self, '_link_extractor'):
```
karlicoss: maybe instead of hasattr, make _link_extractor optional and None if it's not in use?

src/promnesia/sources/reddit.py
```diff
@@ -3,40 +3,153 @@
...
+            title='Reddit save',
+            href=i.url,
+        )
+        yield from self._from_common(i, locator=locator)
+
+
+    # to allow for possible subclassing by the user?
+    def _render_body(self, text: str) -> str:
+        if self.render_markdown:
+            return self._parser_cls(text)._doc_ashtml()
```
karlicoss, line 109: imo a bit nasty when class has different attributes depending on constructor parameters, harder for both people and fo mypy
maybe instead make `_parser_cls` optional and check if it's `None`?

src/promnesia/sources/reddit.py
```diff
@@ -3,40 +3,153 @@
...
+
+    def __init__(self, render_markdown: bool = False):
+
+        try:
+            from .markdown import TextParser, extract_from_text
+            self._link_extractor = extract_from_text
+            self._parser_cls = TextParser
+        except ImportError as import_err:
+            # TODO: add dummy _link_extractor and _parser_cls classes incase
+            # these are called by a subclass?
```
karlicoss, line 64: yeah -- I guess this would also address my other comments below?

src/promnesia/sources/reddit.py
```diff
@@ -3,40 +3,153 @@
...
+                yield Visit(
+                    url=res.url,
+                    dt=i.created,
+                    context=context,
+                    locator=locator,
+                )
+                emitted.add(res.url)
+
+
+def index(*, render_markdown: bool = False, renderer = RedditRenderer) -> Results:
```
karlicoss: hmm, I usually like to keep `index` on the very top (so it's easier to follow what happens on the top level), but here it's got a default parameter, so needs to go after `RedditRenderer`... I guess it's ok

purarue: Can make it optional and put it after the index function instead

src/promnesia/sources/reddit.py
purarue: yeah, makes sense. wanted to do the dummy classes but realized I couldnt subclass the TextParser without it being installed... which means it'd be better to have an abstract type or do some fancy mypy things. seemed like it wouldnt be worth it

karlicoss:
thanks!

PR #238, opened 2021-05-07
FEAT(SRC) new source reading from Signal-Desktop's SQLite

ankostis:
* Ready to be used as described in the docs.
* The new indexer supports the usual `http_only`, `locator_schema` params, plus 2 more specific to Signal desktop's db-file. 
* Testing is absent - unfortunately  needs quite of an effort to do it decently from my sqlciphered db.
* Code should have to be moved/replicated to HPI at some point. 
* Contains a useful & doc-tested `_expand_path()` utility that may worth is weight into `promnesia.common`, even more if it's augmented with `os.path.expandvars()`.

### DIFF
--- a/doc/SOURCES.org
+++ b/doc/SOURCES.org
@@ -141,6 +141,11 @@ for importer, name, ispkg in sorted(pkgutil.walk_packages(
   Uses all local SQLite files found in your Viber Desktop configurations:
   usually in =~/.ViberPC/**/viber.db= (one directory for each telephone number).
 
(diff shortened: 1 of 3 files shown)


ankostis:
Seems to fails a stress-test :-(

karlicoss:
thanks, looks good!
the failure is completely unrelated, seems to happen due to pip version change.. I think I fixed it here https://github.com/karlicoss/promnesia/pull/239

PR #240, opened 2021-05-18
Missing Config Flag for `promnesia config create`

halhenke:
As a Mac User couldnt figure out why it kept installing at `~/Library/Application Support/promnesia` no matter what I did.


karlicoss:
thanks!

purarue:
fwiw, the reason is [appdirs](https://github.com/ActiveState/appdirs/blob/3fe6a83776843a46f20c2e5587afcffe05e03b39/appdirs.py#L187-L190)

PR #242, opened 2021-05-26
FIX signal & report cleared visits

ankostis:
- FIX(signal) undefined var when `http_only` flag is given
- enh(harvesting) log the number of rows cleared (good to know when `--source` given)
- fix(viber) delete leftover `json` import (redundant since sqlite-json extensions were used)


karlicoss:
thanks!

ankostis:
Sorry, discovered somenuances in `dump.py` that has been fixed in this branch b20af12, 
to be squash (if allowed to).

karlicoss:
ah thanks -- I've amended!

PR #252, opened 2021-10-20
sources/html: refactor to allow usage as lib func

purarue:
While [trying to parse HTML fragments of local email files](https://github.com/seanbreckenridge/promnesia/issues/1), wanted to try and use the `sources.html` logic

This refactors the bs4 parsing into a separate function so its usable as a library function - doesn't change any other behavior

### DIFF
--- a/src/promnesia/sources/html.py
+++ b/src/promnesia/sources/html.py
@@ -3,6 +3,7 @@ Extracts links from HTML files
 '''
 
 from pathlib import Path
+from typing import Iterator, Tuple
 from ..common import PathIsh, Visit, Loc, Results, file_mtime
 
 # TODO present error summary in the very end; import errors -- makes sense to show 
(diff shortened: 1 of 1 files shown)


purarue:
Hmm, believe the failure is related to a change in supported CI versions

https://github.com/karlicoss/promnesia/runs/3950405746#step:3:7

karlicoss:
oh yeah just run into the same CI issue

odd, looks like python 3.6 was never officially included in their OSX image? :shrug: https://github.com/actions/virtual-environments/blob/7c673a20aa8f2854dc600e03b3765d733dcc3e2a/images/macos/macos-11-Readme.md#python

I guess not worth investigating, will just exclude it from OSX ci..

ah, so `macos-latest` runner is apparently this https://github.com/actions/virtual-environments/blob/main/images/macos/macos-10.15-Readme.md#python ... but python 3.6 is present in the list. anyway probably easier to just exclude

karlicoss:
thanks, looks good! sorry took me a while to catch up on this :sweat_smile:

PR #256, opened 2021-10-31
reddit: update imports to use match new structure

purarue:
- sources/reddit: update to refelect new HPI format
- check against more specific module not found error
- rename RedditThing to RedditBase

### DIFF
--- a/src/promnesia/sources/reddit.py
+++ b/src/promnesia/sources/reddit.py
@@ -12,10 +12,13 @@ def index(*, render_markdown: bool = False, renderer: Optional['RedditRenderer']
     from . import hpi
     try:
         from my.reddit.all import submissions, comments, saved, upvoted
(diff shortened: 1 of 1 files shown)


purarue:
PR will likely fail since its using pypi HPI version and not the HPI/master, and the changes here were just merged into HPI. See [here](https://github.com/karlicoss/HPI/pull/179#issuecomment-955765079)

Not sure if you'd want me to temporarily uncomment stuff in setup.py to test against HPI/master (but then I'd have to re-add it?, so it would still fail) or if you wanted to do a release once [this is merged in](https://github.com/karlicoss/HPI/pull/184), or something else...

karlicoss:
thanks, looks good! 
I guess would be nice to solve the CI problem 'properly' eventually, but for now I'll just do a new pypi release

Issue #251, opened 2021-10-19
Zotero source


krillin666:
Hi,
Amazing project here !

I saw that you have a source in your HPI for zotero. Would it be possible have it for Promnesia also ? 
It would be really helpful for when opening an url of an article and knowing if it is already in the Zotero database without the friction of having to open Zotero and search for the article !

In your HPI zotero source you have something like:

```python
    item = Item(
        file=Path(path),  # path is a bit misleading... could mean some internal DOM path?
        title=r['title'],
        url=r['url'],
        tags=r['top_tags']
    )
```
Maybe this could be enough to extract the relevant info for Promnesia ? In my opinion just the title or something would be enough to serve the purpose of letting one know if you have a certain url in the database.

Thank you !

karlicoss:
Hey, yeah that would be useful!
And yeah, a basic promnesia module could be very simple, e.g. similar to instapaper https://github.com/karlicoss/promnesia/blob/master/src/promnesia/sources/instapaper.py

It would be also cool to somehow support modules from HPI 'automatically', won't be as nice as a handwritten Promnesia module in general, but at least possible to get some value from the data straightaway

Also if we figure out this https://github.com/zotero/zotero-standalone-build/issues/81, it could also work for opening highlights/books in Zotero directly from Promnesia sidebar

krillin666:
> Hey, yeah that would be useful! And yeah, a basic promnesia module could be very simple, e.g. similar to instapaper [`master`/src/promnesia/sources/instapaper.py](https://github.com/karlicoss/promnesia/blob/master/src/promnesia/sources/instapaper.py?rgh-link-date=2021-10-27T23%3A31%3A36Z)
> 
> It would be also cool to somehow support modules from HPI 'automatically', won't be as nice as a handwritten Promnesia module in general, but at least possible to get some value from the data straightaway
> 


@karlicoss That's amazing and would be really helpful for all academics out here if this is ported to Promnesia.

> Also if we figure out this [zotero/zotero-standalone-build#81](https://github.com/zotero/zotero-standalone-build/issues/81), it could also work for opening highlights/books in Zotero directly from Promnesia sidebar

In relation to this, I think I might be of help here. I was able to make this work in Logseq to open directly Zotero links on the exact page my annotations were made (in the video I enter edit mode first just so you can see that the links in blue are `zotero://open-pdf/library/items....` links):

[...]

Issue #192, opened 2021-02-10
Better integration with Logseq?


karlicoss:
The good new is that Logseq uses org-mode of markdown, so whatever users have in Logseq will work with Promnesia already!

However there might be some Logseq-specific UI enhancements, for example:

* [ ] link to a specific heading/page (like this https://logseq.github.io/#/page/60168f82-f739-4b4b-9852-f91d3d98d314 )
      might require some logseq-specific processing/using the same algorithm Logseq uses to generate IDs
* [ ] link to user's selfhosted Logseq web app
      possible ot hack with `HOOK` config setting
* [x] link to user's Logseq Electron app? 
       similar, but will also require a MIME handler for logseq app
* [ ] maybe some other useful Logseq specific properties?

krillin666:
Any update on this ? 🙂

krillin666:
Do you have any idea to handle Logseq Electron app ?

karlicoss:
hey sorry for late reply!
haven't really had time to look into this, I think best to ask/search on Logseq discord, if you find some clues and share here, would be super helpful :)

krillin666:
Hi @karlicoss, I've got some updates from the Discord chat. It would benefit this project if you could also upvote this feature request: [https://discuss.logseq.com/t/deep-linking-or-url-scheme/3146](https://discuss.logseq.com/t/deep-linking-or-url-scheme/3146).

Keep up the amazing work !

yoyurec:
looks like its possible to clone this https://github.com/karlicoss/promnesia/commit/4f5f7e72b117bf82d0d3b8c388b9cad080b6525e but for [Logseq protocol ](https://github.com/logseq/logseq/pull/4881)(logseq://)  in latest 0.6.6 version 💪

yoyurec:
@krillin666 Logseq protocol added!

yiminglin-ai:
Somehow journal pages don't open correctly. Only standalone pages can work

PR #270, opened 2022-02-10
Automatically use Obsidian URI scheme for Obsidian vaults

Sanqui:
Hi,

Obsidian (a PKM app) has a simple URI scheme to access notes inside vaults (dedicated Obsidian directories): https://help.obsidian.md/Advanced+topics/Using+obsidian+URI

I possibly could've bound Obsidian to your `editor://` thing, but I thought it would be a bit cleaner to automatically use the correct URI scheme (or "handler" as you call it) for detected Obsidian vaults (from the `.obsidian` directory).

Sorry for using the dubious `replacer` functionality, but I tried to integrate a URI scheme through `Options` first and it ended up being a bigger mess...  🙂


src/promnesia/sources/obsidian.py
```diff
@@ -0,0 +1,10 @@
+from ..common import PathIsh
+
+def obsidian_replacer(p: PathIsh) -> str:
+    if not p.startswith("editor://") or not p.endswith('.md'):
```
karlicoss, line 2: this also needs to handle `Path` (since `PathIsh` can be either `Path` or `str`)

src/promnesia/sources/obsidian.py
```diff
@@ -0,0 +1,10 @@
...
+        return p
+    
+    path = p.split('/', 2)[-1]
```
karlicoss, line 5: best to use something like `Path(*Path(p).parts[1:])` -- this would be more portable, e.g. properly work on Windows

karlicoss (changes requested):
Thanks, your change is very welcome! :) 
Hacky, but we can figure out a proper way of handling these later, e.g. it would also be nice for [zotero](https://github.com/karlicoss/promnesia/issues/251) or [logseq](https://github.com/karlicoss/promnesia/issues/192) 

Just one thing, do you think you could add a simple test for this? I don't use obsidian so I won't notice if I manage to break something with refactoring by accident. 
Something similar to [this](https://github.com/karlicoss/promnesia/blob/1b26ccdf9be7c0ac8f8e6e9e4193647450548878/tests/test_auto_indexer.py#L41-L53), and maybe just add a separate directory with small obsidian vault [here](https://github.com/karlicoss/promnesia/tree/master/tests/testdata). Probably best to add a new one instead of changing anything in `auto` dir so you don't have to fix other tests that are using it :)

src/promnesia/sources/obsidian.py
Sanqui: the goal of this is to get of the `editor://` prefix, which is system-independent.  I'm assuming on Windows, `p` will be something like `"editor://C:\Documents and Settings\User\file.md"` in which case this is still fine.  I could also do `p[len("editor://"):]`, sadly in Python these tasks always look a bit iffy.

karlicoss: aah gotcha! yeah good point, I guess it's okay for now, not sure if `editor://` works on Windows anyway

karlicoss:
thanks for the test!
actually realized that replacer doesn't take `PathIsh`, should be string instead: `Replacer = Optional[Callable[[str], str]]`
so just can take `: str` and no need to mess with `Path` -- pushed the fix so we don't bounce the PR back and force

For the future if you'll do more PRs, you can check [this guide](https://github.com/karlicoss/promnesia/blob/master/doc/DEVELOPMENT.org#running-tests--mypy-checks) on running mypy locally -- will make it easier to catch issues and contribute :)

karlicoss (approved):

Sanqui:
Thanks for merging, and my mistake -- I'll make sure to read the guide in the future.

mdroidian:
Do you need the `open?path=` on line 7 in obsidian.py?

I couldn't get it to open the file for me without that.

` uri = f"obsidian://open?path={path}"`

karlicoss:
don't use Obsidian but looks like it? https://help.obsidian.md/Advanced+topics/Using+obsidian+URI#Action+open
@Sanqui can you check if it works for you with `open?path=`? Perhaps you are using an older Obsidian version or something  like that?

Sanqui:
Sorry for the late reply.  The way it's set up now has been working for me.  This way of accessing files is also documented under [Shorthand formats](https://help.obsidian.md/Advanced+topics/Using+obsidian+URI#Shorthand+formats) in the docs.  I'm using latest Obsidian (v0.14.15) and I'm on Linux.  What is your system @mdroidian?

mdroidian:
Windows 10, Obsidian v0.14.15

mdroidian:
> Sorry for the late reply. The way it's set up now has been working for me. This way of accessing files is also documented under [Shorthand formats](https://help.obsidian.md/Advanced+topics/Using+obsidian+URI#Shorthand+formats) in the docs. I'm using latest Obsidian (v0.14.15) and I'm on Linux. What is your system @mdroidian?


Windows 10, Obsidian v0.14.15

karlicoss:
Sorry, just go back to this. So basically, seems that both formats are supported? 
```
obsidian:///absolute/path/to/my note is equivalent to obsidian://open?path=%2Fabsolute%2Fpath%2Fto%2Fmy%20note
```

I'd imagine the `obsidian://open?path=%2Fabsolute%2Fpath%2Fto%2Fmy%20note` is more reliable in general as it's using the URL encoded format. So IMO it would make sense to Obisidian source to use such URLs by default, should work both on Linux and Windows.

PR #275, opened 2022-02-26
org: index URLs in properties

### LINKED ISSUE #218: URLs in PROPERTY drawers not recognized?
Hi!
In the following function to get URLs from an orgmode source: https://github.com/karlicoss/promnesia/blob/8eb9ef3ec30ad045bea7a6a4e89e9579b60e0b1c/src/promnesia/sources/org.py#L103 currently only headings or body are tested for the presence of URLs. The property drawers are only tested for time stamps. I have a setup where notes are captured with a :LINK: property with a URL. Is it possible to add a check to loop over the key-value pairs in the property drawers to test for presence of URLS?
I tried to hack this together, and it works for me for now, but is there a correct way of accomplishing this?
```
    if not n.is_root():        
        try:
            propertydraw = '\n'.join(str(v) for v in n.properties.values()) #ugly hack to get values as single string, can iterate over values?
        except Exception as e:
            logger.exception(e)
            yield e
        else:
            yield from iter_urls(propertydraw, syntax='org')        
```

sopoforic:
This fixes #218, based on the suggestions of @amoghpj and @hrehfeld.

I've eliminated the `if not n.is_root():` test from those since, after all, files can have properties, too.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -96,7 +96,7 @@ DEPS_SOURCES = {
         'mistletoe',
     ],
     ('org'     , 'dependencies for sources.org'     ): [
-        'orgparse>=0.2.0',
+        'orgparse>=0.3.0',
     ],
     ('telegram', 'dependencies for sources.telegram'): [
         'dataset',
--- a/src/promnesia/sources/org.py
+++ b/src/promnesia/sources/org.py
@@ -112,6 +112,15 @@ def iter_org_urls(n: OrgNode) -> Iterator[Res[Url]]:
         yield from iter_urls(heading, syntax='org')
 
     try:
+        values = n.properties.values()
+    except Exception as e:
+        logger.exception(e)
+        yield e
+    else:
+        for v in values:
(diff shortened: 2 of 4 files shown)


sopoforic:
Wait, no, this doesn't work correctly. When I run the test I wrote, which has file-level properties, it seems fine, but when I run it with `promnesia index` I get a lot of `'OrgRootNode' object has no attribute 'properties'`. I'm not sure what's going wrong.

sopoforic:
Ah, my installed `orgparse` was 0.2.4, but the one in the venv I test with was 0.3.1. Apparently `OrgRootNode` didn't have `properties` back then, so it'll be necessary to bump the required version of orgparse to `>=0.3.0`.

karlicoss:
thanks, this looks great! And appreciate the tests :)

karlicoss:
p.s. CI failed but it looks like some random completely unrelated issue with building the extension -- I'll take a look later

hrehfeld:
Thanks so much! :-)

PR #286, opened 2022-04-05
Add source "hackernews"

### LINKED ISSUE #222: Does RSS support only work with Feedbin and Feedly?
Hi, I'd like to keep track of links that I've submitted to Reddit and Hacker News, and I figured I could easily automate that using their RSS feeds. However, I looked at the rss code, and it seems like it only supports Feedly and Feedbin. Is that correct?

telotortium:
Uses the dogsheep module from HPI

Fix https://github.com/karlicoss/HPI/issues/222

### DIFF
--- a/src/promnesia/sources/hackernews.py
+++ b/src/promnesia/sources/hackernews.py
@@ -1,0 +1,41 @@
+'''
+Uses [[https://github.com/karlicoss/HPI][HPI]] dogsheep module to import
+Hacker News items.
+'''
+
+import textwrap
+
+from promnesia.common import Visit, Loc, Results
+
+
+def index() -> Results:
+    from . import hpi
+    from my.hackernews import dogsheep
+
+    for item in dogsheep.items():
+        if isinstance(item, Exception):
+            yield item
+            continue
+        hn_url = item.permalink
+        title = "hackernews"
+        if item.title:
+            title = item.title
+        elif item.text_html:
+            title = item.text_html
+            title = textwrap.shorten(
+                    title, width=79, placeholder="…",
+                    break_long_words=True)
+        # The locator is always the HN story. If the story is a link (as
+        # opposed to a text post), we insert a visit such that the link
+        # will point back to the corresponding HN story.
+        loc = Loc.make(title=title, href=hn_url)
+        urls = [hn_url]
+        if item.url is not None:
+            urls.append(item.url)
+        for url in urls:
+            yield Visit(
+                    url=url,
+                    dt=item.created,
+                    locator=loc,
+                    context=title,
+            )


purarue:
Ah, this is likely failing because dogsheep is such a new module, the module isn't included on the release on pypi (which CI here installs from)

I think @karlicoss usually resolves this by doing a new rolling release on HPI, so CI here works properly

purarue:
Also, could you add [this line](https://github.com/karlicoss/promnesia/blob/6874f2c9b6df1684e8876170bc6caba46e985f9e/src/promnesia/sources/github.py#L13) to the top `index` -- is just some shared code for promnesia sources with use HPI modules. Incase the user doesnt have HPI installed, it warns them

telotortium:
> Also, could you add [this line](https://github.com/karlicoss/promnesia/blob/6874f2c9b6df1684e8876170bc6caba46e985f9e/src/promnesia/sources/github.py#L13) to the top `index` -- is just some shared code for promnesia sources with use HPI modules. Incase the user doesnt have HPI installed, it warns them

Done

src/promnesia/sources/dogsheep.py
```diff
@@ -0,0 +1,28 @@
...
+
+from promnesia.common import Visit, Loc, Results
+
+
+def index() -> Results:
+    from . import hpi
+    from my.hackernews import dogsheep
+
+    for item in dogsheep.items():
+        hn_url = "https://news.ycombinator.com/item?id={item.id}"
```
karlicoss: hmm, so if you look in HPI, the `Item` has `.permalink` attribute, shall we reuse that instead?

src/promnesia/sources/dogsheep.py
```diff
@@ -0,0 +1,28 @@
...
+        title = "hackernews"
+        if item.title:
+            title = item.title
+        elif item.text_html:
+            title = item.text_html
+            if len(title) > 79:
+                title = f"{title[:79]}…"
```
karlicoss: nit: can use [file:///usr/share/doc/python3/html/library/textwrap.html?highlight=textwrap#textwrap.shorten](https://docs.python.org/3/library/textwrap.html?highlight=textwrap#textwrap.shorten) for this

overall, I'd rather prefer such shortening to be implemented in the frontend, since it's more about the presentation rather than indexing. But it's okay for now -- can always change later if we implement it in the extension :)

src/promnesia/sources/dogsheep.py
```diff
@@ -0,0 +1,28 @@
+'''
+Uses [[https://github.com/karlicoss/HPI][HPI]] dogsheep module
+'''
+
+from promnesia.common import Visit, Loc, Results
+
+
+def index() -> Results:
+    from . import hpi
+    from my.hackernews import dogsheep
```
karlicoss: Just fyi: we usually prefer to import from `.all`, e.g. similar to promnesia github module: https://github.com/karlicoss/promnesia/blob/4ca47dd2d3c11e5111416950df7573382f0f007b/src/promnesia/sources/github.py#L14 This allows to use all of the data sources for a particular service instead of just a specific one (e.g. just dogsheep in this case)

You can read more about it following the links here: https://github.com/karlicoss/HPI/blob/master/doc/MODULE_DESIGN.org#an-extendable-module-structure
 
But there isn't one yet in HPI for `my.hackernews`, so it's okay for now. Later we can fix it in a backwards compatible way, e.g. similar to Reddit: 
https://github.com/karlicoss/promnesia/blob/4ca47dd2d3c11e5111416950df7573382f0f007b/src/promnesia/sources/reddit.py#L13-L21

Just leaving it here in case it's useful for you to understand the logic behind Promnesia modules :)

src/promnesia/sources/dogsheep.py
```diff
@@ -0,0 +1,28 @@
...
+        if item.title:
+            title = item.title
+        elif item.text_html:
+            title = item.text_html
+            if len(title) > 79:
+                title = f"{title[:79]}…"
+        loc = Loc.make(
+                title=title, href=(item.url or hn_url))
+        yield Visit(
+                url=hn_url,
```
karlicoss: So I feel like it's a bit suboptimal to define `Loc`/`url` like that here: 

Consider you favorited this item: `https://news.ycombinator.com/item?id=30976119`
- `nh_url` would be `https://news.ycombinator.com/item?id=30976119`
- `item.url` would be `https://www.wired.co.uk/article/europe-police-facial-recognition-prum`

Now when you visit `https://www.wired.co.uk/article/europe-police-facial-recognition-prum`, you won't get any Promnesia visits at all (because the server is matching by `url`).
When you visit `https://news.ycombinator.com/item?id=30976119`, you'll get a Promnesia visit pointing at `https://www.wired.co.uk/article/europe-police-facial-recognition-prum` as permalink, which is kind of useless since the Hackernews article will have this link itself.

imo it makes more sense to do something like this instead: 

```
    loc = Loc.make(title=title, href=hn_url)
    urls = [hn_url]
    iurl = item.url
    if iurl is not None:
        urls.append(iurl)
    
    for url in urls:
        yield Visit(
            url=url,
            dt=item.created,
            locator=loc,
            context=title,
        )
```
that way you'll get a Promnesia visit when you visit both urls, and it will always point to the hackernews article (so you'll get a 'reverse lookup' from the article on Wired site)

karlicoss (changes requested):
hey @telotortium , thanks for the contribution! Sorry just got to check this out.

Agree it would be better to rename this to `hackernews.py` instead. Maybe we'll have a `dogsheep` module one day, but it'd probably need to talk to Dogsheep directly instead and be more or less service agnostic (this one is just reusing Dogsheep code for data retrieval)

l also left a few comments -- fairly minor, and I'm happy to merge as it is, but if you have some time to fix, would be grateful!

Also yeah @seanbreckenridge is right about the CI failing -- I've released a new HPI version and rerun. Would be nice to solve it somehow so it works against master as well, but going to be tedious... P.S. it still fails with mypy issues -- so @telotortium let me know if you need help resolving

src/promnesia/sources/dogsheep.py
telotortium: Thanks, that's good to know. I suppose in order to do this for hackernews, we would have to modify dogsheep and materialistic in HPI to have them both export the same format for each item.

src/promnesia/sources/dogsheep.py
telotortium: Good point - I've implemented this now.

src/promnesia/sources/dogsheep.py
karlicoss: Yep, something like this -- it's not always possible, but usually there is a reasonable common denominator (which we put in `common.py`), and then it can be inherited by both specific implementations. 
So if someone cares about merged data they can query it from `.all` submodule, whereas if they'd rather prefer richer data (e.g. API data usually has more fields than GDPR exports), they can use a more specific submodule.

src/promnesia/sources/dogsheep.py
karlicoss: great, but think you forgot to push :)

telotortium (commented):
Please review and approve.

telotortium:
Now that you pointed out how `Visit` is constructed, and what the difference is between `url` and `locator` (by the way, you should make that more clear in the comments on the `Visit` class), I'm going to try to make a change upstream to add the story each comment was posted on to the Dogsheep database, so that we can use that info here: https://github.com/dogsheep/hacker-news-to-sqlite/issues/4. However, it's probably best to merge this first.

src/promnesia/sources/hackernews.py
```diff
@@ -0,0 +1,38 @@
...
+import textwrap
+
+from promnesia.common import Visit, Loc, Results
+
+
+def index() -> Results:
+    from . import hpi
+    from my.hackernews import dogsheep
+
+    for item in dogsheep.items():
```
karlicoss, line 15: so the return type of `dogsheep.items` is `Iterator[Res[Item]]`, where `Res` basically means it can be an `Exception` or `Item`, this is for [flexible error handling](https://beepb00p.xyz/mypy-error-handling.html#kiss)
Basically you want to add here something like
```
        if isinstance(item, Exception):
            yield item
            continue
```
e.g. similar to https://github.com/karlicoss/promnesia/blob/4ca47dd2d3c11e5111416950df7573382f0f007b/src/promnesia/sources/github.py#L25

karlicoss (changes requested):
Looks good, thanks! 

CI failed with mypy, I suggested a change that should help. Btw here's how you can run type checks locally https://github.com/karlicoss/promnesia/blob/master/doc/DEVELOPMENT.org#running-tests--mypy-checks

Also some errors are irrelevant (related to fbmessengerexport/twitter), not sure how it happened, but not your fault -- I'll take a look at the meantime

karlicoss:
thanks for your work & fixes!

PR #298, opened 2022-04-23
Automatically use Logseq URI scheme for Logseq graph

yoyurec:
Do same as Obsidian https://github.com/karlicoss/promnesia/pull/270, but for Logseq protocol (https://github.com/logseq/logseq/pull/4881)


src/promnesia/sources/logseq.py
```diff
@@ -0,0 +1,12 @@
+import os.path
+
+def logseq_replacer(path: str, root: str) -> str:
+    if not path.startswith("editor://") or not path.endswith('.md'):
```
karlicoss: I guess for logseq also `.org` is possible? probably worth including here

src/promnesia/sources/auto.py
```diff
@@ -185,6 +186,10 @@ def index(
                 logger.debug('detected %s as Obsidian vault', root)
                 replacer = obsidian_replacer
 
+        if root is not None and (root / "logseq").exists():
```
karlicoss: a minor thing, but maybe worth detecting "logseq/config.edn"? Since currently this would match a `logseq` code repository, which is not we want here

karlicoss (changes requested):
looks good, appreciate your contribution! :) 
some minor comments + mypy failed, likely need to add a similar check for `root is not None` and pass `str(root)` to the replacer

src/promnesia/sources/auto.py
yoyurec: sounds good

src/promnesia/sources/logseq.py
yoyurec: yepp

yoyurec:
BTW, Logseq displays text urls nice and automake them clickable:
image

Why MD files are parsed for links only `[title](link)` syntax, not just simple text `http://blah.com`?
As user i'm sure i've mentioned link in PKM, expect to see it in sidebar results, but Promnesia skip it on indexing 😕

yoyurec:
ping )

karlicoss (approved):
Hey, was away :) 
Thanks, looks good! I fixed the remaining mypy issue myself, just so we can merge it and don't have to go back and forth.
It usually helps a lot to run mypy locally, e.g. with `tox -e mypy-core` and `tox -e mypy-misc`, you can read more about it here https://github.com/karlicoss/promnesia/blob/master/doc/DEVELOPMENT.org#running-tests--mypy-checks

karlicoss:
thanks, merged!

> Why MD files are parsed for links only [title](link) syntax, not just simple text http://blah.com?

Hmm, personally I have very little markdown, so not sure, but I wonder if this bit to blame

https://github.com/karlicoss/promnesia/blob/13727a514266b4c27cdf89d887d514abe2be4362/src/promnesia/sources/markdown.py#L53-L56

Basically, the markdown parser detects `[]()` urls as a proper block (`Link`), but if it's just a `http` link in text, it doesn't treat it any special. Perhaps we should also handle `mistletoe.RawText` blocks, and call `extract_urls` on them, that should probably catch these links? 

P.S. btw you'll be very welcome to join and participate in https://memex.zulipchat.com
(no need to register, you can use github login)

PR #300, opened 2022-04-26
Add Logseq and Obsidian sources docs

yoyurec:
Users had no idea that Promnesia works just perfect with Logseq & Obsidian.
Fixed )


yoyurec:
Hm... Errors https://github.com/karlicoss/promnesia/runs/6175945407?check_suite_focus=true

karlicoss:
thanks! 
hmm the CI error seems transient, some NPM registry issues, hopefully it'll just disappear after a bit

PR #301, opened 2022-04-26
Logseq file path encode, Auto source helpers rename

yoyurec:
Encode special chars in file path


doc/config.py
```diff
@@ -14,6 +14,10 @@
 # - structured formats like JSON and CSV
 from promnesia.sources import auto
 
+# Logseq protocol support for auto indexer
+from promnesia.sources import logseq
```
karlicoss: don't think it's actually needed here? it only contains helper functions for `auto` module, no need to import it in the config

yoyurec: right. thought i've added config but forgot inport for it )) but it's not a source it's a replacer

yoyurec: little confusing. i think as it's not THE SOURCE, but just replacer ("helper") - it should be renamed to `auto_logseq` & `auto_obsidian` or even put them to subfolder `auto`

karlicoss: yep, good idea!

yoyurec: fixed!

PR #302, opened 2022-04-27
Sidebar more css classes and styles

yoyurec:
NEW RP here - https://github.com/karlicoss/promnesia/pull/308


yoyurec:
...you just need to up version in configs

PR #307, opened 2022-04-30
Extension: Fix multiple inits (onload)

### LINKED ISSUE #177: extension: make more resilient against constant page updates
Some sites (e.g. twitter.com) emit onUpdated all the time when scrolled, which causes constant requests to the backend and redrawing the sidebar.
Might be nice to debounce it or something.

### LINKED ISSUE #295: Chrome extension sidebar sometimes closes right after clicking the eye icon
Symptoms: clicking the eye icon, usually on a page where I actually have visits, seems to do nothing at first. If I click repeatedly and rapidly on the eye icon, I can see the sidebar flicker briefly into existence, but then is closed within milliseconds after opening.

Don't know yet how to reproduce - seems intermittent. Often if I refresh the page, the sidebar will stay open when I click on it.

yoyurec:
Fixes https://github.com/karlicoss/promnesia/issues/295
Fixes https://github.com/karlicoss/promnesia/issues/177

> Please be aware `chrome.tabs.onUpdated` will also fired for iframes, if a page contains many iframes, each completed iframe will trigger the event though you have checked changeInfo.status. Use `chrome.webNavigation.onCompleted`

### DIFF
--- a/extension/src/background.js
+++ b/extension/src/background.js
@@ -99,9 +99,9 @@ function getIconStyle(result: Result): IconStyle {
 }
 
 
-async function updateState (tab: chrome$Tab) {
+async function onTabLoaded(tab) {
     const url = unwrap(tab.url);
-    const tabId = unwrap(tab.id);
+    const tabId = unwrap(tab.tabId);
 
     if (ignored(url)) {
         // todo reflect in the sidebar/popup?
@@ -214,7 +214,7 @@ async function updateState (tab: chrome$Tab) {
         visits = new Visits(url, url, [visits])
     }
     console.assert(visits instanceof Visits)
-  
+
     // right, we can't inject code into error pages (effectively, internal). For these, display popup instead of sidebar?
     // TODO and show system wide notification instead of tab notification?
     // https://stackoverflow.com/questions/32761782/can-a-chrome-extension-run-code-on-a-chrome-error-page-i-e-err-internet-disco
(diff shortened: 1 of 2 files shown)


yoyurec:
`Cannot get chrome.webNavigation because property webNavigation is missing in object type [1]. [prop-missing]
node_modules/flow-interfaces-chrome/interfaces/Chrome.js`

em... ) 
too old package
https://github.com/Shraymonks/flow-interfaces-chrome/search?q=webNavigation

extension/src/background.js
```diff
@@ -99,9 +99,9 @@ function getIconStyle(result: Result): IconStyle {
 }
 
 
-async function updateState (tab: chrome$Tab) {
+async function onTabLoaded(tab) {
     const url = unwrap(tab.url);
-    const tabId = unwrap(tab.id);
+    const tabId = unwrap(tab.tabId);
```
karlicoss, line 110: hmm, looks like this is causing these for me
 
```
requesting! https://news.ycombinator.com/item?id=31224707
common.js:19 Uncaught (in promise) Error: undefined or null!
    at unwrap (common.js:19:11)
    at shouldProcessPage (background.js:583:62)
    at openSidebar (background.js:702:24)
    at eval (background.js:248:22)
```

extension/src/background.js
```diff
@@ -453,41 +453,25 @@ chrome.webNavigation.onDOMContentLoaded.addListener(detail => {
...
-
-    // TODO vvvv these might need to be cleaned up; not sure how relevant...
-    // page refresh: loading -> complete (no url at any point)
-    // clicking on link: loading (url) -> complete
-    // opening new link: loading -> loading (url) -> complete
-    // ugh. looks like 'complete' is the most realiable???
-    // but, I checked with 'complete' and sometimes it would reload many things with loading -> complete..... shit.
 
-    // also if you, say, go to web.telegram.org it's gonna show multiple notifications due to redirect... but perhaps this can just be suppressed..
+    if (info.frameId !== 0) {
```
karlicoss, line 474: should this check go first (before the URL check)? would make it faster and less logging

extension/src/background.js
```diff
@@ -453,41 +453,25 @@ chrome.webNavigation.onDOMContentLoaded.addListener(detail => {
...
 // chrome.tabs.onReplaced.addListener(updateState);
 
-// $FlowFixMe
-chrome.tabs.onUpdated.addListener(defensify(async (tabId: number, info, tab: chrome$Tab) => {
-    // too spammy in logs
-    delete tab.favIconUrl
-    delete info.favIconUrl
-    //
-    console.debug('onUpdated %o %o', tab, info)
+// $FlowExpectedError
```
karlicoss: yeah, looks like they are incomplete for these  https://github.com/Shraymonks/flow-interfaces-chrome/issues/1#issue-145335503

karlicoss (commented):
So I had an error in the log, in particular, 'always show sidebar' stoppedworking
So I think this requires more changes since we switched from the Tab type to oncompleted callback -- I committed Flow types to make the errors more obvious

yoyurec:
any updates with checking functionality here?

karlicoss:
ah have you seen CI logs for `build_extension` pipeline? (not sure you have the permissions) -- otherwise you can try running `npm run flow` locally (just make sure to pull my latest commits, where I annotated the function properly)

I think there are more changes that need to be done -- the object we are now passing into `onTabLoaded` [here](https://github.com/karlicoss/promnesia/blob/f076f1c374f71b754881e0e33b2eaf42f0dd729e/extension/src/background.js#L108) 

Haven't checked exactly what's missing though -- it may be possible it's just a few occasions of replacing `.id` with `.tabId` like here? https://github.com/karlicoss/promnesia/blob/f076f1c374f71b754881e0e33b2eaf42f0dd729e/extension/src/background.js#L529-L530

yoyurec:
so what's the status? will this PR be used somehow or you have found another solution?

karlicoss:
Hey, @yoyurec  sorry for late reply -- as I mention in the comment https://github.com/karlicoss/promnesia/pull/307#issuecomment-1120473302 , there are still some outstanding issues -- but otherwise I haven't look at it

karlicoss:
closing in favor of https://github.com/karlicoss/promnesia/pull/354 (thanks for the inspiration though :) )

PR #308, opened 2022-05-01
Sidebar more css classes and styles

### LINKED ISSUE #102: extension: sidebar issues on certain sites (e.g. truncation/scrolling/etc)
Feel free to comment here if you find more such examples

- [ ] https://arthurperret.fr/2020/05/22/ecrire-et-editer/ (@bepolymathe)
      same for Hypothesis, so potentially site markup is somehow unusual
- [ ] https://www.quaternum.net (@bepolymathe)
      possibly something to do with Jekyll?
- [ ] https://psychonautwiki.org/wiki/Cannabidiol (sidebar detaches when scrolled)
- [ ] https://www.bloomberg.com/europe (might be behind popups) https://github.com/karlicoss/promnesia/issues/146#issue-727850381

yoyurec:
* Fixes https://github.com/karlicoss/promnesia/issues/102
* Added more classes to markup for deeper restyling possibilities
* Header "sticky", only visits scrolled
* Added '.button' styles
* Main buttons more recognisible - with emoji icons
* Added badges counter for sources filter
* Added `active` status for "Mark visited" button
* Added `active` status for selected source filter (!)
* Reworked a lot of styling for `#visits` elements
   * bookmark source items with icon ⭐
   * box colors
   * timestamp monospaced font
 * Text/bg/border colors use CSS variables

Before:   

After:
image

### DIFF
--- a/extension/src/options.js
+++ b/extension/src/options.js
@@ -143,8 +143,60 @@ function defaultOptions(): StoredOptions {
         position_css: `
 /* you can use devtools to find other CSS attributes you can tweak */
 
+/* Tweak colors */
+
+/* Common used colors, you are welcome to customize them! */
+/*
+.promnesia {
+    --opacity: 1;
+    --main: rgba(255, 254, 252, var(--opacity));
+    --secondary: #e8eef0;
+    --thirdly: #f3f4f6;
+    --borders: #dddddd;
+    --text: #222222;
+    --text-secondary: #999999;
+    --accent: 178, 102, 40;
+    --accent-secondary: #2aa198;
+}
+*/
+
+/* Styling of main elements. Mostly you don't need to customize it */
+/*
+.promnesia {
+    --header-bg: var(--secondary);
+    --source-text: rgb(var(--accent));
+    --count-text: var(--thirdly);
+    --visits-item-bg: var(--thirdly);
+    --visits-item-border: var(--borders);
+    --locator-text: var(--accent-secondary);
+    --timestamp-text: var(--text-secondary);
+    --date-bg: var(--main);
+    --date-color: var(--text);
+}
+*/
+
+/* Custom fonts */
+/*
+@import url('https://fonts.googleapis.com/css2?family=Fira+Code:wght@400;600&family=PT+Serif:ital,wght@0,400;0,700;1,400;1,700&display=swap');
+.promnesia {
+    font-family: "PT Serif", serif;
+    font-size: 1em;
+}
+*/
+
+
+/* shadows example */
+/*
+#promnesia-sidebar-filters {
+    text-transform: capitalize;
+    background: linear-gradient(to bottom, var(--thirdly) 0%, var(--main) 100%);
+}
+*/
+
+
 /* tweak sidebar position/size/style */
-#promnesia-sidebar {
+#promnesia-frame {
+
     /* you can also use
        --left/--top/--bottom
        to change the sidebar position */
@@ -154,14 +206,7 @@ function defaultOptions(): StoredOptions {
 
     /* you can also use any other valid CSS
        easiest is to experiment in devtools first */
-    background-color: rgba(236, 236, 236, 0.8);
 }
-
-/* tweak elements within the sidebar */
-#promnesia-sidebar .src {
-    font-weight: bold;
-}
-
 
 /* tweak 'visited' marks: specify hex color here */
 :root {
--- a/extension/src/search.html
+++ b/extension/src/search.html
@@ -3,12 +3,12 @@
     <head>
         <meta charset="UTF-8">
         <title>Search</title>
-        <link rel="stylesheet" type="text/css" href="sidebar.css"> 
+        <link rel="stylesheet" type="text/css" href="sidebar.css">
     </head>
-    <body>
+    <body id='promnesia-search' class='promnesia'>
         <form id='search_id'>
         <input type='text' id='query_id' size='80' autofocus>
-        <button type='submit'>Search</button>
+        <button class='button' type='submit'>🔎 Search</button>
         </form>
         <script src='search.js'/></script>
 
--- a/extension/src/sidebar-outer.css
+++ b/extension/src/sidebar-outer.css
@@ -1,7 +1,22 @@
-#promnesia-sidebar.promnesia {
+/* Resets for sites  https://github.com/karlicoss/promnesia/issues/102 */
+body.promnesia-sidebar-active {
+    transform: none;
+}
+.promnesia-sidebar-active #promnesia-frame {
+    margin: 0;
+    padding: 0;
+    max-width: none;
+    max-height: none;
+    border: 0;
+    border-left: 1px solid #999;
+}
+
+#promnesia-frame {
     /* e.g. left should result in: left initial, right 0px, top 0px, bot 0px, height 100%, width 30%
        need to set irrelevant dimensions to 0px, otherwise it misbehaves
     */
+    --right: 1;
+    --size: 30%;
 
     --is-v: var(--left, var(--right ));
     --is-h: var(--top , var(--bottom));
@@ -23,12 +38,12 @@
     height: var(--hhelper, 100%);
 
     position: fixed;
-    z-index: 9999;
+    z-index: 2147483647;
 
     display: none;
 }
 
-body.promnesia {
+.promnesia {
     padding-left  : calc(var(--left)   * var(--size));
     padding-right : calc(var(--right)  * var(--size));
     padding-top   : calc(var(--top)    * var(--size));
--- a/extension/src/sidebar.css
+++ b/extension/src/sidebar.css
@@ -1,6 +1,25 @@
-.promnesia body {
-    font-size: 15px;
-    color: black;
+.promnesia {
+    /* Common used colors, you are welcome to customize them! */
+    --opacity: 1;
+    --main: rgba(255, 254, 252, var(--opacity));
+    --secondary: #e8eef0;
+    --thirdly: #f3f4f6;
+    --borders: #dddddd;
+    --text: #222222;
+    --text-secondary: #999999;
+    --accent: 178, 102, 40;
+    --accent-secondary: #2aa198;
+
+    /* Styling of main elements. Mostly you don't need to customize it */
+    --header-bg: var(--secondary);
+    --source-text: rgb(var(--accent));
+    --count-text: var(--thirdly);
+    --visits-item-bg: var(--thirdly);
+    --visits-item-border: var(--borders);
+    --locator-text: var(--accent-secondary);
+    --timestamp-text: var(--text-secondary);
+    --date-bg: var(--main);
+    --date-color: var(--text);
 }
 
 .promnesia .error {
@@ -14,34 +33,71 @@
     text-decoration: none;
 }
 
-/* todo would be nice to properly reuse it and share with the 'mark visited' */
-.promnesia .src {
-    /* color: orange; /\* TODO monospace font or smth? *\/ */
-    color: #b26628;
-
-    margin-right: 0.2em;
-    white-space: nowrap;
-}
-
 a.promnesia-visited {
     outline: 0.5em solid;
 }
 
-.src-filter .src {
-    cursor: pointer;
-    display: inline-block;
+#promnesia-sidebar {
+    display: flex;
+    flex-flow: column;
+    height: 100%;
+    margin: 0;
+    color: var(--text);
+    background: var(--main) !important;
+    overscroll-behavior: contain;
 }
 
-.src-filter .src:hover {
-    color: red; /* TODO ?? */
+#promnesia-sidebar-toolbar {
+    display: flex;
+    padding: 6px 8px 0 8px;
+    border-bottom: 1px solid var(--borders);
+    overflow: hidden;
+    background: var(--header-bg);
+}
+
+#promnesia-sidebar-toolbar #button-mark.active {
+    border-color: rgb(var(--accent));
+    box-shadow: rgb(var(--accent)) 0px 1px 2px 0px;
+}
+
+#promnesia-sidebar-toolbar #button-close {
+    margin-left: auto;
+    padding: 0;
+    width: 34px;
+    height: 34px;
+    border-radius: 50%;
+    font-size: 0.7em;
+}
+
+#promnesia-sidebar-filters {
+    padding: 10px 10px 0px 10px;
+    background: linear-gradient(to bottom, var(--thirdly) 0%, var(--main) 100%);
+}
+
+#promnesia-sidebar-filters .src.button {
+    color: var(--source-text);
+    padding: 4px 8px;
+    font-weight: 400;
+    box-shadow: none;
+}
+
+#promnesia-sidebar-filters .button:hover,
+#promnesia-sidebar-filters .button.active {
+    border-color: rgb(var(--accent));
+    box-shadow: rgb(var(--accent), 0.5) 0 1px 1px;
+}
+
+#promnesia-sidebar-container {
+    flex-grow: 1;
+    overflow-y: auto;
 }
 
 /* meh, but ok for now */
 #visits {
-    /* make table occupy whole sidebar */
-    width: 100%;
+    list-style: none;
+    margin: 8px;
+    padding: 0;
 }
-
 
 #visits .context {
     white-space: pre-wrap; /* need to keep linebreaks  */
@@ -51,6 +107,11 @@ a.promnesia-visited {
 /* TODO use proper flex layout... */
 #visits .locator {
     padding-left: 1em;
+    color: var(--locator-text);
+}
+
+#visits .locator a {
+    color: var(--locator-text);
 }
 
 #visits .index {
@@ -58,21 +119,21 @@ a.promnesia-visited {
     display: inline-block;
 }
 
-
-#visits {
-    list-style: none;
-    margin: 0;
-    padding: 0;
+#visits .normalised_url {
+    padding-left: 1em;
 }
 
 #visits li {
-    margin-top: 1em;
+    margin: 0;
+    padding: 6px;
+    background: var(--visits-item-bg);
+    border: 1px solid var(--visits-item-border);
+    border-radius: 3px;
+    overflow-wrap: anywhere;
+}
 
-    padding: 0.2em;
-
-    background: #e6e6e6cc;
-    border: 3px solid;
-    border-color: transparent;
+#visits li+li {
+    margin-top: 8px;
 }
 
 #visits li #relative_indicator {
@@ -90,7 +151,7 @@ a.promnesia-visited {
 }
 
 #visits li:hover {
-    border-color: #ddd;
+    border-color: rgba(var(--accent), 0.5);
 }
 
 #visits li.relative:hover {
@@ -104,25 +165,97 @@ a.promnesia-visited {
 #visits .datetime {
     white-space: nowrap;
     float: right;
-
-    color: #666;
-
     cursor: pointer;
+    font-family: monospace;
+    color: var(--timestamp-text);
 }
-
 
 /* TODO not great, but quickest way to render it neatly */
 /* TODO not sure which order users expect first? */
 #visits .time {
-    /* color: blue; */
-    margin-right: 1em;
+    padding-right: 1em;
 }
+
 #visits .date {
-    /* color: red; */
     width: 11ch; /* TODO not sure how to alighn them properly across all of boxes... */
     display: inline-block; /* otherwise width doesn't work on span */
     text-align: end;
-
+    border-radius: 6px;
+    padding: 2px 4px;
+    background: var(--date-bg);
 }
 
+#visits .src {
+    color: var(--source-text);
+    font-weight: 700;
+}
 
+#visits .src.bookmark::before {
+    content: "\2B50";
+    margin-right: 4px;
+    font-size: 0.9em;
+    line-height: 1.2em;
+}
+
+#visits .src.logseq::before {
+    content: "\1F4D4";
+    margin-right: 4px;
+    font-size: 0.9em;
+    line-height: 1.2em;
+}
+
+#promnesia-sidebar-filters .src-count {
+    background-color: rgb(var(--accent));
+    border-radius: 9999px;
+    display: inline-flex;
+    align-items: center;
+    justify-content: center;
+    font-weight: 400;
+    font-size: 0.75rem;
+    line-height: 1;
+    padding: 0.25rem 0.5rem;
+    color: var(--count-text);
+    margin-left: 1ex;
+}
+
+/* Buttons */
+.promnesia .button {
+    color: var(--text);
+    background-color: var(--main);
+    border: 1px solid var(--text-secondary);
+    border-radius: 6px;
+    box-shadow: var(--borders) 0 1px 1px;
+    cursor: pointer;
+    display: inline-block;
+    /* set font for better emoji support */
+    font-family: -apple-system, system-ui, "Segoe UI", Helvetica, Arial, sans-serif, "Apple Color Emoji", "Segoe UI Emoji";
+    font-size: 14px;
+    font-weight: 500;
+    line-height: 20px;
+    list-style: none;
+    padding: 6px 12px;
+    margin: 0 6px 6px 0;
+    appearance: none;
+    user-select: none;
+    -webkit-user-select: none;
+    vertical-align: middle;
+}
+
+.promnesia .button:hover {
+    border-color: var(--text);
+}
+
+.promnesia .button:focus {
+    outline: 1px transparent;
+}
+
+/* Search page*/
+#promnesia-search #search_id {
+    display: flex;
+}
+
+#promnesia-search #query_id {
+    font-size: 1em;
+    height: 1.7em;
+    margin-right: 0.5em;
+}
--- a/extension/src/sidebar.js
+++ b/extension/src/sidebar.js
@@ -15,8 +15,13 @@ function get_or_default(obj, key, def) {
     return res === undefined ? def : res;
 }
 
-
-const SIDEBAR_ID   = 'promnesia-sidebar';
+const PROMNESIA_FRAME_ID   = 'promnesia-frame';
+const PROMNESIA_CLASS   = 'promnesia';
+const SIDEBAR_ID = 'promnesia-sidebar';
+const HEADER_ID = 'promnesia-sidebar-header';
+const FILTERS_ID = 'promnesia-sidebar-filters';
+const FILTERS_CLASS = 'src-filter';
+const TOOLBAR_ID = 'promnesia-sidebar-toolbar';
 const CONTAINER_ID = 'promnesia-sidebar-container';
 
 
@@ -30,7 +35,7 @@ const Cls = {
 }
 
 
-const SIDEBAR_ACTIVE = 'promnesia';
+const SIDEBAR_ACTIVE = 'promnesia-sidebar-active';
 
 const doc = document;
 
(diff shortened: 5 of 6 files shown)


extension/webpack.config.js
```diff
@@ -10,7 +10,7 @@ const T = {
 };
 
 const env = {
-    TARGET : process.env.TARGET,
+    TARGET : process.env.TARGET || T.CHROME,
```
karlicoss: we should leave it out, would be nice to figure out why the ./build script not working. Might result in some nasty bugs in tests etc..

extension/src/options.js
```diff
@@ -141,10 +141,43 @@ function defaultOptions(): StoredOptions {
         // TODO add docs on positioning
         // TODO eh, would be nice to make it work with --right: true. right now it doesn't
         position_css: `
+		@import url('https://fonts.googleapis.com/css2?family=Fira+Code:wght@400;600&family=PT+Serif:ital,wght@0,400;0,700;1,400;1,700&display=swap');
```
karlicoss: should be wrapped in a comment?

extension/src/sidebar.css
```diff
@@ -1,6 +1,48 @@
-.promnesia body {
-    font-size: 15px;
-    color: black;
+@import url('https://fonts.googleapis.com/css2?family=Fira+Code:wght@400;600&family=PT+Serif:ital,wght@0,400;0,700;1,400;1,700&display=swap');
```
karlicoss: I'd rather avoid loading random fonts as the default (esp from Google) -- let's leave it up to the user settings :)

extension/src/sidebar.css
```diff
@@ -104,25 +132,157 @@ a.promnesia-visited {
...
+}
+
+#visits .src.logseq::before {
+    content: "\1F4D4";
+    margin-right: 4px;
+    font-size: 0.9em;
+    line-height: 1.2em;
+}
+
+#promnesia-sidebar-filters .src-num {
```
karlicoss: best to name it .src-count (or scr-total).
'number' usually refers to the index rather than total count

extension/src/sidebar.js
```diff
@@ -344,7 +368,9 @@ async function* _bindSidebarData(response: Visits) {
     const sidebar = new Sidebar(opts);
 
     const cont = await sidebar.getContainer();
-    await sidebar.clear(); // TODO probably, unnecessary?
+    const filtersEl = await sidebar.getFiltersEl();
+    filtersEl.innerHTML = '';
```
karlicoss: we should use `safeSetInnerHTML` function (otherwise it might trip up some Mozilla security checks)

extension/src/sidebar-outer.css
```diff
@@ -23,12 +25,12 @@
     height: var(--hhelper, 100%);
 
     position: fixed;
-    z-index: 9999;
+    z-index: 2147483647;
```
karlicoss, line 41: hmm, is this max int? is there a reason to set it like that?

extension/src/sidebar.css
```diff
@@ -104,25 +132,157 @@ a.promnesia-visited {
...
 
+#promnesia-sidebar-toolbar {
+    padding: 6px 8px 0 8px;
+    border-bottom: 1px solid var(--bg);
+    overflow: hidden;
+    background: var(--header-bg);
+}
+
+#promnesia-sidebar-toolbar .button.active {
+    border-color: var(--source-color);
```
karlicoss: IMO it's way too subtle -- maybe should set some thickness as well?

extension/src/sidebar.js
```diff
@@ -68,39 +78,54 @@ class Sidebar {
...
+			search_button.title = "Search links in database";
+            sidebar_toolbar.appendChild(search_button);
         }
         {
             // TODO only on mobile?
-            const elem = cdoc.createElement('button');
-            elem.appendChild(cdoc.createTextNode('Close'));
-            elem.addEventListener('click', defensify(async () => {
+            const close_button = cdoc.createElement('button');
+			close_button.classList.add('button', 'button-close');
```
karlicoss: maybe instead of classes, use button-close/button-search etc as `id`? since they are unique elements

extension/src/sidebar.css
```diff
@@ -104,25 +132,157 @@ a.promnesia-visited {
...
+}
+
+#promnesia-sidebar-toolbar .button-close {
+    float: right;
+    border-radius: 999px;
+    padding: 4px;
```
karlicoss: image
hmm I think this padding makes it misaligned with the other buttons -- is it necessary?

extension/src/sidebar.css
```diff
@@ -104,25 +132,157 @@ a.promnesia-visited {
...
+}
+
+#promnesia-sidebar-container {
+    flex-grow: 1;
+    overflow-y: auto;
+}
+
+#promnesia-sidebar-filters {
+    text-transform: capitalize;
```
karlicoss: hmm, not sure what it's supposed to do, but it doesnt look like it capitalizes anything for me

extension/src/sidebar.js
```diff
@@ -152,7 +177,7 @@ class Sidebar {
     }
 
     getFrame(): ?HTMLIFrameElement {
-        return ((doc.getElementById(SIDEBAR_ID): any): ?HTMLIFrameElement);
+        return ((doc.getElementById(PROMNESIA_FRAME_ID): any): ?HTMLIFrameElement);
```
karlicoss, line 183: TODO container id changed?

extension/src/sidebar.css
```diff
@@ -104,25 +132,157 @@ a.promnesia-visited {
...
+    color: var(--color);
+    background-color: var(--bg);
+    border: 1px solid var(--bg4);
+    border-radius: 6px;
+    box-shadow: rgba(27, 31, 35, 0.04) 0 1px 0,
+        rgba(255, 255, 255, 0.25) 0 1px 0 inset;
+    box-sizing: border-box;
+    cursor: pointer;
+    display: inline-block;
+    font-family: -apple-system, system-ui, "Segoe UI", Helvetica, Arial,
```
karlicoss: imo default font is fine, let's just keep it up to the configuration?

extension/src/sidebar.css
```diff
@@ -104,25 +132,157 @@ a.promnesia-visited {
...
+    padding: 6px 12px;
+    margin: 0 6px 6px 0;
+    position: relative;
+    appearance: none;
+    user-select: none;
+    -webkit-user-select: none;
+    touch-action: manipulation;
+    vertical-align: middle;
+    white-space: nowrap;
+    word-wrap: break-word;
```
karlicoss: hmm, do we need word-wrap if we use nowrap above?

extension/src/sidebar.css
```diff
@@ -104,25 +132,157 @@ a.promnesia-visited {
...
+        sans-serif, "Apple Color Emoji", "Segoe UI Emoji";
+    font-size: 14px;
+    font-weight: 500;
+    line-height: 20px;
+    list-style: none;
+    padding: 6px 12px;
+    margin: 0 6px 6px 0;
+    position: relative;
+    appearance: none;
+    user-select: none;
```
karlicoss, line 239: hmm, why? I get there is probably little point to select text there, but I'd prefer to keep css as small as possible :)

extension/src/sidebar.css
```diff
@@ -104,25 +132,157 @@ a.promnesia-visited {
...
+    font-family: -apple-system, system-ui, "Segoe UI", Helvetica, Arial,
+        sans-serif, "Apple Color Emoji", "Segoe UI Emoji";
+    font-size: 14px;
+    font-weight: 500;
+    line-height: 20px;
+    list-style: none;
+    padding: 6px 12px;
+    margin: 0 6px 6px 0;
+    position: relative;
+    appearance: none;
```
karlicoss, line 238: hm, why? doesn't seem to have any effect for me

extension/src/sidebar.css
```diff
@@ -104,25 +132,157 @@ a.promnesia-visited {
...
+    display: inline-block;
+    font-family: -apple-system, system-ui, "Segoe UI", Helvetica, Arial,
+        sans-serif, "Apple Color Emoji", "Segoe UI Emoji";
+    font-size: 14px;
+    font-weight: 500;
+    line-height: 20px;
+    list-style: none;
+    padding: 6px 12px;
+    margin: 0 6px 6px 0;
+    position: relative;
```
karlicoss: what's that for? doesn't seem to impact anything for me

extension/src/sidebar.css
```diff
@@ -104,25 +132,157 @@ a.promnesia-visited {
...
+
+/* Buttons */
+.promnesia .button {
+    color: var(--color);
+    background-color: var(--bg);
+    border: 1px solid var(--bg4);
+    border-radius: 6px;
+    box-shadow: rgba(27, 31, 35, 0.04) 0 1px 0,
+        rgba(255, 255, 255, 0.25) 0 1px 0 inset;
+    box-sizing: border-box;
```
karlicoss: box-sizing/box-shadow don't seem to impact anything for me?

extension/src/sidebar.css
```diff
@@ -104,25 +132,157 @@ a.promnesia-visited {
...
+    font-weight: 500;
+    line-height: 20px;
+    list-style: none;
+    padding: 6px 12px;
+    margin: 0 6px 6px 0;
+    position: relative;
+    appearance: none;
+    user-select: none;
+    -webkit-user-select: none;
+    touch-action: manipulation;
```
karlicoss: I guess it seems useful, just wondering how did you notice that it's necessary? i.e. seems to only have effect on the phone?

extension/src/search.html
```diff
@@ -3,12 +3,12 @@
...
         <title>Search</title>
-        <link rel="stylesheet" type="text/css" href="sidebar.css"> 
+        <link rel="stylesheet" type="text/css" href="sidebar.css">
     </head>
-    <body>
+    <body id='promnesia-search' class='promnesia'>
         <form id='search_id'>
         <input type='text' id='query_id' size='80' autofocus>
-        <button type='submit'>Search</button>
+        <button class='button' type='submit'>🔎 Search</button>
```
karlicoss, line 11: hmm, why remove type=submit?

extension/src/sidebar.css
```diff
@@ -104,25 +132,157 @@ a.promnesia-visited {
...
     text-align: end;
+    border-radius: 6px;
+    padding: 2px 4px;
+    color: var(--color2);
+    background: var(--date-bg);
+}
 
+#promnesia-sidebar {
+    display: flex;
+    flex-flow: column;
```
karlicoss: hmm, are display: flex and flex-flow necessary? it behaves the same without them

extension/src/sidebar.css
```diff
@@ -104,25 +132,157 @@ a.promnesia-visited {
...
+    height: 100%;
+    margin: 0;
+    background: var(--bg);
+    overscroll-behavior: contain;
```
karlicoss: hmm, what's it for here?
I usually try to write comments for obscure css properties so it's easier for people to understand why it's here, so it's not lost during refactoring

extension/src/sidebar.css
```diff
@@ -104,25 +132,157 @@ a.promnesia-visited {
...
+    font-family: monospace;
 }
 
-
 /* TODO not great, but quickest way to render it neatly */
 /* TODO not sure which order users expect first? */
 #visits .time {
-    /* color: blue; */
-    margin-right: 1em;
+    padding-right: 0.5em;
```
karlicoss: hmm, why split it into margin and padding?

extension/src/sidebar.css
```diff
@@ -104,25 +132,157 @@ a.promnesia-visited {
...
+    -webkit-user-select: none;
+    touch-action: manipulation;
+    vertical-align: middle;
+    white-space: nowrap;
+    word-wrap: break-word;
+}
+
+.promnesia .button:hover {
+    border-color: var(--color2);
+    text-decoration: none;
```
karlicoss: hmm, what does it do?

extension/src/sidebar.css
```diff
@@ -104,25 +132,157 @@ a.promnesia-visited {
...
+    background-color: var(--source-color);
+    border-radius: 9999px;
+    display: inline-flex;
+    align-items: center;
+    justify-content: center;
+    font-weight: 400;
+    font-size: 0.75rem;
+    line-height: 1;
+    padding: 0.25rem 0.5rem;
+    color: var(--num-color);
```
karlicoss: similar to above, renameto --source-count-color?

extension/src/options.js
```diff
@@ -141,10 +141,43 @@ function defaultOptions(): StoredOptions {
...
+
 /* you can use devtools to find other CSS attributes you can tweak */
 
+/* Teweak colors */
+
+/* Ligt mode */
```
karlicoss: nit: typos -- tweak/light

extension/src/sidebar.css
```diff
@@ -104,25 +132,157 @@ a.promnesia-visited {
...
+
+#promnesia-sidebar-container {
+    flex-grow: 1;
+    overflow-y: auto;
+}
+
+#promnesia-sidebar-filters {
+    text-transform: capitalize;
+    padding: 6px 10px 0px 10px;
+    background: linear-gradient(to bottom, var(--bg3) 0%, var(--bg) 100%);
```
karlicoss: imo kinda subjective, so maybe keep in the private config? but don't mind it either

extension/src/sidebar.css
```diff
@@ -1,6 +1,48 @@
...
+
+/* Ligt mode */
+.promnesia {
+    --opacity: 1;
+    --bg: rgba(255, 254, 252, var(--opacity));
+    --bg1: #e8eef0;
+    --bg2: #f3f4f6;
+    --bg3: #ddd;
+    --bg4: #ccc;
+    --color: #222;
```
karlicoss: hmm, these names are really not descriptive, can we maybe add an explanation/link to the screenshot explaining what are these? 
(fine to do later once we settle on the final version)

extension/src/sidebar.css
```diff
@@ -1,6 +1,48 @@
...
+    --source-color: rgb(var(--accent-color));
+    --source-bg: var(--bg2);
+    --num-color: var(--bg2);
+    --visits-item-bg: var(--bg2);
+    --locator-color: var(--accent-color2);
+    --date-bg: var(--bg);
+    --date-color: var(--color);
+    --time-color: var(--color2);
+
+    font-family: "PT Serif", serif;
```
karlicoss: hmm imo it's subjective -- shall we keep it in your private config?

extension/src/sidebar.css
```diff
@@ -1,6 +1,48 @@
...
+    font-size: 1em;
```
karlicoss: isn't it kinda pointless to have 1em as the font size? it doesn't do anything right?

extension/src/sidebar-outer.css
```diff
@@ -1,7 +1,20 @@
-#promnesia-sidebar.promnesia {
+/* Resets for sites */
+body.promnesia-sidebar-active {
+    transform: none;
+}
+.promnesia-sidebar-active #promnesia-frame {
+    margin: 0;
+    padding: 0;
+    max-width: none;
```
karlicoss, line 8: hmm what's it for?

extension/src/options.js
yoyurec: No, sidebar uses it )
Importing font implemented in iframe scope (as it should), but some sites (for. ex. github) have warnings, fonts not loaded:
`Refused to load the stylesheet 'https://fonts.googleapis.com/css2?family=Fira+Code:wght@400;600&family=PT+Serif:ital,wght@0,400;0,700;1,400;1,700&display=swap' because it violates the following Content Security Policy directive: "style-src 'unsafe-inline' github.githubassets.com". Note that 'style-src-elem' was not explicitly set, so 'style-src' is used as a fallback.`

So font duplicated in options, b'cose their css scope = initial page scope, no policy warnings.

extension/src/sidebar.css
yoyurec: ok, will remove it and put to my own options settings )
anyway it looks almost the same, but little narrower

extension/src/sidebar-outer.css
```diff
@@ -1,7 +1,20 @@
-#promnesia-sidebar.promnesia {
+/* Resets for sites */
+body.promnesia-sidebar-active {
+    transform: none;
```
karlicoss, line 3: oh now I see why it's here https://github.com/karlicoss/promnesia/issues/102#issuecomment-1114237352
can you add a link to the issue please? because otherwise it's not obvious what this code is for :)

extension/src/sidebar.css
yoyurec: ok ) 
hm... initally i want to name it `src-count` 🤔

extension/src/sidebar.js
yoyurec: smart!

extension/webpack.config.js
yoyurec: you mean delete?

extension/src/sidebar-outer.css
yoyurec: yes, this is max z-index. will beat all popups, cookie blocks, `scroll to top` arrows, etc...
reason is clear and simple - some selements bleads throung sidebar ((
reason same as before, but 9999 isn't big enougth

extension/src/sidebar.css
yoyurec: give me the number )

extension/src/sidebar.js
yoyurec: ok

extension/src/sidebar.css
yoyurec: will make it beautiful

extension/src/sidebar.css
yoyurec: it capitalize word `filter: ` (BTW, why not "Filter:" in initial js?) + source buttons....
can you put screen?

extension/src/sidebar.js
yoyurec: TODO? how to see it?

extension/src/sidebar.css
yoyurec: default font have no proper emoji icons ((
windows:
image

extension/src/sidebar-outer.css
karlicoss: aah gotcha -- didn't expect there would be any pages with such z-inde

extension/src/sidebar.css
yoyurec: common button ui behavior. sometimes users do clicking and text selects. but it shouldn't

extension/src/sidebar.js
karlicoss: ah sorry this was todo for me -- forgot to remove it, this bit is good :)

extension/webpack.config.js
karlicoss: yeah, you could keep it locally but not commit?

extension/src/sidebar.css
yoyurec: also common buttons stuff.
removing default platrform styling...
https://css-tricks.com/almanac/properties/a/appearance/
something like reset.css... maybe mostly for mobiles...

extension/src/sidebar.css
yoyurec: dunno. copypasted from button generator. meybe used in some complicated content case
will delete

extension/src/sidebar.css
yoyurec: shadows impacts, look closer ))

box-sizing can be removed.

extension/src/sidebar.css
yoyurec: can delete. code from generator

extension/src/search.html
yoyurec: is it removed?

extension/src/sidebar.css
yoyurec: default is row
image

extension/src/sidebar.css
yoyurec: when init page have scroll. and our sidebar have scroll. when you scroll with mouse wheel inside sidebar and reach its end, but countinue to scroll with mouse wheel - whole init page will countinue to scroll. it's by default and it's annoying. 
` overscroll-behavior: contain;` disables it.

extension/src/sidebar.css
yoyurec: for better look in case of adding border (was in my previous revisions). will delete

extension/src/sidebar.css
yoyurec: from generator. will delete.

extension/src/sidebar.css
yoyurec: right

extension/src/options.js
yoyurec: tnx. it was deep **niht** 😴

extension/src/sidebar.css
yoyurec: ok

extension/src/sidebar.css
yoyurec: yes

extension/src/sidebar.css
yoyurec: for default font it's not need, is i see...

extension/src/sidebar-outer.css
yoyurec: for issue https://github.com/karlicoss/promnesia/issues/102

extension/src/sidebar-outer.css
yoyurec: ok

extension/src/search.html
karlicoss: :man_facepalming:  sorry, not sure where I looked

extension/webpack.config.js
yoyurec: ok

extension/src/sidebar.css
karlicoss: yep, but is `display: flex` necessary at all?

extension/src/sidebar.css
yoyurec: it's different )
dunno. will delete both

karlicoss: the `white-space: nowrap` is definitely useful, you can see it if you have a long tag with whitespace, and try a narrower screen in the devtools. Just not sure about the break-word :)

extension/src/sidebar.css
karlicoss: ah, fair enough -- can you add a comment about this please :)

extension/src/sidebar.css
yoyurec: it's "2nd layer" of variables ))
most important selectors has own good named vars:
image

but inside them i also using vars... THE THEME vars, their names here a little not speackable.
they reused in different properties, so can't be named `---border-color` as it's also used in shadow, etc...

it's like abstract vars, and then below more linked to blocks..

extension/src/sidebar.css
yoyurec: done

extension/src/sidebar.css
yoyurec: > IMO it's way too subtle

you can use your own options config ahahahahahah )))

yoyurec: i tried 2px. it's tooooo big and ugly (((

extension/src/sidebar.css
karlicoss: ah -- I think it didn't work for some reason, but seem that you removed it now anyway
Rither way, imo if the user wanted to capitalize source names they would just name them from the capital letter. Also it makes it a bit misleading for debugging etc since the source names in data would be different from what they appear as.
renaming to `Filter: ` is fine though, I just have a very relaxed attitude towards capital letters xD

extension/src/sidebar.css
karlicoss: ok! Not a big deal, so fine to keep as it is if you prefer :) 
could also use coloring change or something like that, but I don't mind to keep it as is

extension/src/sidebar.css
```diff
@@ -1,6 +1,43 @@
-.promnesia body {
-    font-size: 15px;
-    color: black;
+/* Ligt mode */
```
karlicoss: typo :)

extension/src/sidebar.css
karlicoss: hm, okay makes sense

extension/src/sidebar.css
yoyurec: like that?
image

extension/src/sidebar.css
yoyurec: at first i  want to name them `ligh`, `lighter`, etc... but then i introduced dark mode and in became little confusing.
will try something like `primary`, `secondary`... or....
at least sinplify (again and again 😁) amount of used colors more...

extension/src/sidebar.css
karlicoss: yup, looks good!

extension/src/sidebar.js
```diff
@@ -344,7 +371,9 @@ async function* _bindSidebarData(response: Visits) {
     const sidebar = new Sidebar(opts);
 
     const cont = await sidebar.getContainer();
-    await sidebar.clear(); // TODO probably, unnecessary?
+    const filtersEl = await sidebar.getFiltersEl();
+    safeSetInnerHTML(filtersEl, '');
```
yoyurec: em... it doesn't clean ((
filters populated `onCompleted`, i need to clean them. previos code did it well...
image

karlicoss: ah, that's odd! I think we should add cleaning to `safeSetInnerHTML`, must have missed it by accident

extension/src/sidebar.css
yoyurec: done!

extension/src/sidebar.css
yoyurec: tnx

extension/src/sidebar.js
yoyurec: i think we can use native API https://developer.mozilla.org/en-US/docs/Web/API/Element/replaceChildren

extension/src/sidebar.css
yoyurec: yes, it allow to have scroll for #visits. 
overflow-y: auto; + height:100% not working correct.

extension/src/sidebar.css
yoyurec: i'll make it lighter, but it do visually beautify a lot )

karlicoss: yeah it's fine to keep here, don't worry

extension/src/sidebar.js
karlicoss: looks like JS flow is complaining, but I think it's just because `replaceChildren` was added relatively recently, in 2020
https://developer.mozilla.org/en-US/docs/Web/API/Element#browser_compatibility
I think it should be fine just to suppress its error in this case, 2020 is quite old in browser versions
Or even better maybe wrap it in try/catch just in case (with a comment). Worst case the extension will be slighly broken for the user, but not completely :)

yoyurec:
phew... looks like all comments + few new spotted things fixed!

extension/src/sidebar.js
```diff
@@ -427,10 +455,23 @@ async function* _bindSidebarData(response: Visits) {
...
+			// set active status to filter buttons
+			// 1) reset status from all sources
+			filtersEl.childNodes.forEach(filter_element => {
+				if (filter_element.nodeType == Node.ELEMENT_NODE) {
+					filter_element = ((filter_element: any): HTMLElement);
+					filter_element.classList.remove('active');
+				}
+			});
+			// 2) set active to clicked source
+			tag_c.classList.add('active');
```
karlicoss: code seems misaligned in this block, can you shift it one space to the left?

extension/webpack.config.js
karlicoss: hopefully just this left :)

karlicoss (changes requested):

extension/webpack.config.js
yoyurec: gone )

extension/src/sidebar.js
yoyurec: done

karlicoss (approved):
awesome, thanks so much! :tada: 
will merge once CI completes

PR #313, opened 2022-05-02
Use sqlcipher3 instead of pysqlcypher3

krillin666:
Just a heads up. For user on version 3.10 of python, the signal source will not work by default because it uses pysqlcipher. The pysqlcipher3 module does not work with Python 3.10, solved this by installing sqlcipher3 ( https://pypi.org/project/sqlcipher3/ ) and substitute the import on line 340:


karlicoss:
ah, thanks!
I think promnesia messed up the filename somehow, if you look closely it renamed it -- can you try pushing from a local git? 
I should really figure out what's causing it and fix it, really annoying

krillin666:
> ah, thanks! I think promnesia messed up the filename somehow, if you look closely it renamed it -- can you try pushing from a local git? I should really figure out what's causing it and fix it, really annoying

Sorry but I have really no experience with github, could you point me on how can I push from local ? Do I need to fork you repo ?

karlicoss:
image

Right -- I reproduced the issue -- seems that the marks Promnesia adds somehow mess it up.
As a workaround -- can you do the same change again, but before pressing "Commit changes/Propose changes", disable the visited marks? ('Mark/Unmark visited urls' in the context menu or `Ctrl-Alt-V` shortcut if you're using Firefox)
I'll try to fix the actual extension bug in the meantime

PR #316, opened 2022-05-04
Use sqlcipher3 instead of pysqlcypher3

krillin666:
Just a heads-up. For user on version 3.10 of python, the signal source will not work by default because it uses pysqlcipher. The pysqlcipher3 module does not work with Python 3.10, solved this by installing sqlcipher3 ( https://pypi.org/project/sqlcipher3/ ) and substitute the import on line 340:


karlicoss:
hmm, annoying, seems it's renamed again. I committed the chance to fixup the file name, will squash it before merging
created an issue to track the bug which causes broken names during github inline changes, will try to fix soonish https://github.com/karlicoss/promnesia/issues/317

thanks for the change again!

PR #327, opened 2022-06-05
deprecate old browser history export script

karlicoss:
https://github.com/seanbreckenridge/browserexport should be used instead


ankostis:
But how?
(filed new #339 about it)

PR #333, opened 2022-08-26
Pocket improvements

### LINKED ISSUE #294: Pocket: improve locator_title and context fields in Promnesia.
The JSON from https://github.com/karlicoss/pockexport contains `resolved_title` and `excerpt` fields, which could be used for `locator_title` and `context` respectively. Also, add canonicalization to Pocket URLs so that https://app.getpocket.com/read/ is canonicalized to https://getpocket.com/read/ - right now, the Promnesia extension doesn't even recognize that I've visited the same Pocket page before.

telotortium:
- Set `locator_title` and `context` from `pockexport` data.
- Add a backlink `Visit` with `url` set to the Pocket URL and `locator`
  set to the original URL.
- Canonicalize `getpocket.com` to `app.getpocket.com`.

Fix https://github.com/karlicoss/promnesia/issues/294

### DIFF
--- a/src/promnesia/cannon.py
+++ b/src/promnesia/cannon.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python3
 """
 Experimental libarary for normalising domain names and urls to achieve 'canonical' urls for content.
-E.g. 
+E.g.
  https://mobile.twitter.com/demarionunn/status/928409560548769792
 and
  https://twitter.com/demarionunn/status/928409560548769792
@@ -55,6 +55,9 @@ dom_subst = [
     ('np.reddit.'     , 'reddit.'),
 
     ('m.facebook.'    , 'facebook.'),
+    # app.getpocket.com is the canonical domain in the JSON returned by
+    # https://github.com/karlicoss/pockexport, so let's canonicalize to that.
+    ('getpocket.'     , 'app.getpocket.'),
 ]
 
 def canonify_domain(dom: str) -> str:
@@ -626,6 +629,14 @@ FB_PATTERNS = [
     r'F/notes/U/P',
 ]
 
+PKP = r'^(app)?\.getpocket\.com'
+
+PK_PATTERNS = [
+    {
+        'ID': r'\d+',
+    },
+    PKP + '/read/ID',
+]
 
 # NOTE: right, I think this is just for analysis so far... not actually used
 PATTERNS = {
@@ -636,6 +647,7 @@ PATTERNS = {
     'stackoverflow': SO_PATTERNS,
     'facebook'  : FB_PATTERNS,
     'wikipedia' : WK_PATTERNS,
+    'pocket'    : PK_PATTERNS,
     # 'news.ycombinator.com': YC_PATTERNS,
 }
 
--- a/src/promnesia/sources/pocket.py
+++ b/src/promnesia/sources/pocket.py
@@ -7,16 +7,22 @@ from ..common import Visit, Loc, Results
 def index() -> Results:
     from . import hpi
     from my.pocket import articles
+
     # TODO use docstring from my. module? E.g. describing which pocket format is expected
 
     for a in articles():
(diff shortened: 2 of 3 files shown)


telotortium:
@karlicoss Seems like the test configuration is broken.

src/promnesia/sources/pocket.py
```diff
@@ -7,22 +7,43 @@
...
+        # Add a reverse locator so that the Promnesia browser extension shows a
+        # link on the Pocket page back to the original URL.
+        loc_rev = Loc.make(title=title, href=a.url)
         hls = a.highlights
         if len(hls) == 0:
             yield Visit(
                 url=a.url,
                 dt=a.added,
-                context=None,
+                context=a.json.get('excerpt', None),
```
karlicoss: nit: let's extract `excerpt =` variable to avoid duplication?

src/promnesia/sources/pocket.py
```diff
@@ -7,22 +7,43 @@
...
                 locator=loc,
             )
+            # Add a visit with the Pocket URL so that we can see that we have
```
karlicoss: Hmm, I had thought about adding it at first, but then thought it is pointless because when you're on Pocket website, you can see the original URL in the interface. Do you think this adds some value? I don't mind really, just curious what I'm missing because a similar pattern is also applicable to other data sources like Instapaper :)

src/promnesia/sources/pocket.py
```diff
@@ -7,22 +7,43 @@
 def index() -> Results:
     from . import hpi
     from my.pocket import articles
+
     # TODO use docstring from my. module? E.g. describing which pocket format is expected
 
     for a in articles():
-        loc = Loc.make(title='pocket', href=a.pocket_link)
+        title = a.json.get('resolved_title', 'pocket')
```
karlicoss: Yeah thanks, I guess this makes sense. 
Although sometimes `resolved_title` is `""` in my Pocket export. There is also `given_title` which is not empty for such entries, so maybe worth trying to use that too while we're at it?

src/promnesia/sources/pocket.py
```diff
@@ -7,22 +7,43 @@
...
         for hl in hls:
             yield Visit(
                 url=a.url,
                 dt=hl.created,
                 context=hl.text,
                 locator=loc,
             )
+            # Add a visit with the Pocket URL so that we can see that we have
+            # visited before in Promnesia browser extension.
+            yield Visit(
```
karlicoss: same as above -- not sure there is much point adding all these visits for each highlight when you're on Pocket website anyway and can see them in Pocket interface? But again don't think it's a big deal either way so curious to know what you think

src/promnesia/cannon.py
```diff
@@ -55,6 +55,9 @@ def try_cutr(suffix: str, s: str) -> str:
     ('np.reddit.'     , 'reddit.'),
 
     ('m.facebook.'    , 'facebook.'),
+    # app.getpocket.com is the canonical domain in the JSON returned by
+    # https://github.com/karlicoss/pockexport, so let's canonicalize to that.
+    ('getpocket.'     , 'app.getpocket.'),
```
karlicoss, line 60: :+1: thanks!

karlicoss (changes requested):
Hey, sorry, was on holiday and only just got to this :see_no_evil:  
This looks good, just some minor comments.
I looked at the `end2end` CI pipeline -- seems like some internal github actions issue, I'll take a look later but feel free to ignore it for now, your changes shouldn't impact these tests

src/promnesia/sources/pocket.py
telotortium: Looking at this, I agree with you - it is pretty useless. I'll remove it.

src/promnesia/sources/pocket.py
telotortium: Done

karlicoss:
sorry, only just got to this! looks great, thanks so much for your changes :heart:

Issue #341, opened 2022-12-30
Running Promnesia crashes Facebook (and some other sites)


Stvad:
Recipe is something like:
- Go to FB page that has Promnesia matches
- After promnesia loads - scroll timeline
- 💥

--- 
A similar story occurs on my digital garden website: https://vlad.roam.garden. Algorithm is similar but instead of scrolling timeline - the action is to open internal ink (which should open in a new side-panel). 

The common thing between those two, I think is, triggering a React re-render

Running it in Firefox 109.0b6, macOS 13.1

karlicoss:
Yeah, can confirm it happens on your garden site.
I think the issue is caused by "mark as visited" feature (it says 'experimental' because I sort of anticipated it might break stuff). Can you confirm it doesn't crash when "mark as visited" is disabled?

To workaround there are a few options at the moment
- disable Promnesia on "vlad.roam.garden" altogether (sad, but presumably on your own garden it's not as useful anyway since you remember more context)
- disable "mark as visited" by default and only use it via hotkey (a bit sad)
- use "exclude multiple links from mark as visited (element zapper)" in promneisa menu on the site and select the main column of your site -- at least it worked for me. Also somewhat sad option because it means links to your garden won't be marked when linked from other sites

I think the right thing to do (considering it's inevitable that page modifications would break something) would be:
- separate setting to disable "mark as visited" feature for a domain/regex/exact page. Exactly same semantics as normal blocklist for Promnesia, just a separate one. Possibly same for highlights feature as well? 
- the zapper should block DOM path (similar to the way ad blockers work) rather than using the URLS inside that dom path

Stvad:
Can confirm that disabling `Mark visited` prevents crashes both on roam.garden and FB. 

Will think on what workaround works best 🤔

Proper solutions wise: I think it's possible to make the feature work more reliably without crashing things. 

For example my extension https://github.com/transclude-me/extension does very similar things mechanics-wise (mark up relevant links & show popup when you hover over them) without causing any crashes (afaik)

I believe what makes the difference is that I do mark-up in CSS-only way (add a class to relevant links and style them differently) and don't modify the DOM

karlicoss:
Interesting, thanks for the reference! (cool extension by the way :) )

IIRC the main reason I was wrapping elements into extra HTML was to add a bit which activates the popup on hover -- not sure if it's possible with CSS? 

[...]

PR #367, opened 2023-01-29
use tippy.js for visited marks

karlicoss:
see https://github.com/karlicoss/promnesia/issues/341


yoyurec:
why not NanoPop? ))
https://github.com/Simonwep/nanopop

image

\* running away 🏃‍♂️ ))

Issue #365, opened 2023-01-29
extension/indexers: ensure less 'span' in visited marks


karlicoss:
for @karlicoss a good example is this page https://github.com/settings/billing

![image](https://user-images.githubusercontent.com/291333/215339536-0ba9aee1-da44-4ac5-a23c-8e06f36fb6fb.png)

One thing that would be nice is to make google takout visit 'boring' -- that can be achieved by removing contexts in promnesia indexer (for simple visits they only contain the domain name) https://github.com/karlicoss/promnesia/blob/master/src/promnesia/sources/takeout.py#L39

Another is perhaps adding an option to hide "boring" visits completely, or even better to hide certain sources. Maybe ideally with a javascript-based hook to start with

karlicoss:
after: 
![image](https://user-images.githubusercontent.com/291333/215647946-a6488745-5345-489f-8e09-9dec2d729b5a.png)

still could benefit from excluding them from the visited marks UI I guess but it's quite a bit better

AnweshGangula:
Just sharing a related concern I'm facing:

The context in below screenshot renders the HTML text from Logseq

![](https://github.com/karlicoss/promnesia/assets/22996896/fbc1bc06-c7ee-42a5-83dd-8bd362127801)

> I have raised a related issue with contexts in `search.html` in #430

PR #369, opened 2023-01-31
sources.takeout: clean contexts for Activity events form useless entries

karlicoss:
this makes it less spammy when showing visited marks/in the sidebar they mostly should be treated like 'boring' link visits

see https://github.com/karlicoss/promnesia/issues/365


karlicoss:
@seanbreckenridge not sure if you're using promnesia/takeout source these days, but tagging you in case you have an opinion :)

purarue:
[I do, yep](https://github.com/seanbreckenridge/dotfiles/blob/9d1f71418a0e61f3cd907cab629e73bf7fd60e9d/.config/promnesia/config.py#L4) -- looks good

We had talked about this somewhere in the past, not sure where -- cant really even search for it in promnesia because me searching for 'Visited' to find the conversation just brings up :weary: 

image

karlicoss:
uhoh -- one day I'll get to implement pagination or just use a proper indexer :sweat_smile:

PR #370, opened 2023-02-03
fix typo

raphwriter:
(Dockerfile seems to work without the (misspelled) directory anyhow)


karlicoss:
ah, seems that if WORKDIR doesn't exist, docker will create it, that explains why it worked. thanks!

PR #371, opened 2023-02-03
add dependencies to Dockerfiles

raphwriter:
had to add these dependencies in order for the docker images to run

### DIFF
--- a/docker/docker_files/Dockerfile
+++ b/docker/docker_files/Dockerfile
@@ -11,7 +11,7 @@ COPY setup.py /usr/src/
 
 RUN pip install --no-cache-dir more_itertools pytz sqlalchemy cachew \
                 appdirs urlextract python-magic \
-                tzlocal hug \
+                tzlocal \
                 logzero HPI beautifulsoup4 lxml mistletoe orgparse dataset fastapi uvicorn
 
 ENV PPATH=/usr/src/promnesia:${PPATH}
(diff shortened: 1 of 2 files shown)


docker/docker_files/Dockerfile
```diff
@@ -12,7 +12,7 @@ COPY setup.py /usr/src/
 RUN pip install --no-cache-dir more_itertools pytz sqlalchemy cachew \
                 appdirs urlextract python-magic \
                 tzlocal hug \
```
karlicoss: could you remove `hug` as well please?

docker/docker_files/Dockerfile-indexer
```diff
@@ -1,5 +1,6 @@
 FROM promnesia:latest
 
 RUN apt-get update && apt-get install -y cron
+RUN pip install fastapi uvicorn
```
karlicoss: hmm, I don't think it's necessary since this Dockerfile is only indexing (e.g. it didn't have `hug` dependency previously either)

karlicoss (changes requested):
Thanks! Yeah this file is contributed by someone else, so I didn't notice it breaking
Would be nice to integrate it into CI somehow, maybe even a basic test is enough, since the rest is well tested through other means

docker/docker_files/Dockerfile-indexer
raphwriter: makes sense -- sorry :)

docker/docker_files/Dockerfile
raphwriter: Imade the changes, hope I did it right... I'm not that familiar with how github pull requests work (yet)

karlicoss:
thanks!

PR #374, opened 2023-02-04
takeout: ignore placevisit events

purarue:
this might `ImportError` for people if they have not upgraded `google_takeout_parser`... should a warning be added there?

### DIFF
--- a/src/promnesia/sources/takeout.py
+++ b/src/promnesia/sources/takeout.py
@@ -1,7 +1,7 @@
 '''
 Uses HPI [[https://github.com/karlicoss/HPI/blob/master/doc/MODULES.org#mygoogletakeoutpaths][google.takeout]] module
 '''
-from typing import Iterable, Set, Type
+from typing import Iterable, Set, Any
 import warnings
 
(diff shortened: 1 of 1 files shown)


src/promnesia/sources/takeout.py
```diff
@@ -13,7 +13,7 @@ def index() -> Results:
 
     try:
         from my.google.takeout.parser import events
-        from google_takeout_parser.models import Activity, YoutubeComment, LikedYoutubeVideo, ChromeHistory, PlayStoreAppInstall, Location
+        from google_takeout_parser.models import Activity, YoutubeComment, LikedYoutubeVideo, ChromeHistory, PlayStoreAppInstall, Location, PlaceVisit
```
karlicoss: Yeah you're right about ImportError. Maybe we should make it `_seen: Set[str]` instead and use class names there instead of actual types? Then the import won't even be necessary (along with PlayStoreAppInstall and Location)

purarue:
Somewhat related: https://github.com/karlicoss/HPI/issues/266

karlicoss:
thanks, looks good!

Issue #339, opened 2022-11-25
doc: browser-history instructions missing


ankostis:
I noticed that _browser-history_ instructions are completely absent from `SOURCES.org`, so I could provide a MR (if i figured how to do it:-)).

Would it suffice to do the following:

1. run some this *browserexport* commands (which ones??) 

2. configuring [the respective HPI MODULE](https://github.com/karlicoss/HPI/blob/master/doc/MODULES.org#mybrowser), and 

3. somehow _(?)_ instructing *promnesia* to hook from the above HPI SOURCE. 

Another Q regards the old txt files, generated by the deprecated-script: how to also parse them into a unified browser history?

purarue:
For 1, you should use `browserexport save -b [browser] -t ~/data/browsing`  or something like that, see [here](https://github.com/seanbreckenridge/browserexport#save)

Can copy over the example part of the readme:

```
$ browserexport save -b firefox --to ~/data/browser_history
$ browserexport save -b chrome --to ~/data/browser_history
$ browserexport save -b safari --to ~/data/browser_history
```

For 3, after `my.browser` is configured in HPI, the Visit source just needs a quick transformation, like [this](https://github.com/seanbreckenridge/promnesia/blob/master/promnesia_sean/sources/browsing.py). That source (currently in my own promnesia module repo) can be copied over/merged here

karlicoss:
Yeah, I've been meaning to switch promnesia to use https://github.com/seanbreckenridge/promnesia/blob/master/promnesia_sean/sources/browsing.py
Just will need to implement a fallback to the old method (in case someone doesn't have HPI configured, similar to what takeout does)
Also will run some final comparisons first just in case nothing is lost/timezone issues

purarue:
@karlicoss if you'd like, I can make a PR with some instructions for using browserexport here. Should that go in `SOURCES.org` or somewhere else? edit: perhaps a section in `GUIDE.org`?

karlicoss:
@seanbreckenridge thanks, of course would be appreciated!  `GUIDE.org` feels a bit more generic/high level. I think a separate section in `SOURCES.org` would be good (since the ones that are listed there are just autogenerated from docstrings).

ankostis:
Given the opportunity, browsers forget their history pretty soon, and extending the history span is the most precious artifact of promnesiq for me.  I would like the instructions to explain in detail how not to lose one accumulated history elements, when updating. And whether duplicates and overlaps are dropped.

[...]

PR #375, opened 2023-02-06
sources.browser: switch to use HPI + seanbreckenridge/browserexport

karlicoss:
later will rename to browser, and implement defensive fallback onto browser_legacy

adapted from https://github.com/seanbreckenridge/promnesia/blob/master/promnesia_sean/sources/browsing.py

related: https://github.com/karlicoss/promnesia/issues/339

Old vs new modules produce almost identical results (tested on all of my chrome & firefox databases). The only minor differences are:

- old database timestamps end with +00:00 UTC, new ones with +00:00 -- likely because browserexport is using timezone.utc instead of pytz
- previously locator was pointing at the database file, now it's pointing at the URL

  I guess it's not necessarily in the 'spirit' of locator field, but on the other hand, not that it's very useful to point to an sqlite file either.
  Perhaps later it could be in some sort of extra debug field instead.


karlicoss:
@seanbreckenridge just fyi -- your module worked great, only had to add durations :)

karlicoss:
Heh, didn't believe I'd make this `tmp_config` thing work for a fallback, but it does :grinning:

purarue:
Switch was trivial :+1: , index looks good

https://github.com/seanbreckenridge/dotfiles/commit/f8c6866253d19ff781ca113b55b6bd427c6872bb

PR #386, opened 2023-04-15
sources.smscalls: propogate errors if there are breaking schema changes

purarue:
related to: https://github.com/karlicoss/HPI/pull/289

Am unsure if CI here will fail till the return type from HPI is updated, but I think the change is small enough should be able to merge after publishing a release there...

Just to show works locally with new types:

```
$ mypy -p promnesia.sources.smscalls
Success: no issues found in 1 source file

$ hpi doctor -S promnesia.sources.smscalls
✅ OK  : promnesia.sources.smscalls                        
✅     - stats: {'index': {'index': {'count': 371, 'last': datetime.datetime(2023, 4, 14, 20, 6, 18, 459000, tzinfo=datetime.timezone.utc)}}}
```

### DIFF
--- a/src/promnesia/sources/smscalls.py
+++ b/src/promnesia/sources/smscalls.py
@@ -11,11 +11,18 @@ def index() -> Results:
 
     for m in messages():
 
+        if isinstance(m, Exception):
+            yield m
+            continue
+
         urls = extract_urls(m.message)
         if len(urls) == 0:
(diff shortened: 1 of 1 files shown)


karlicoss:
I think mypy can complain about unnecessary None checks, but it's a custom flag which we don't have :)

PR #387, opened 2023-04-17
Update donation link for org-mode

unendiverse:
the orgmode.org link is dead. the org-mode.org website provides this link as their donation link.


karlicoss:
thanks!

Issue #398, opened 2023-06-11
unable to build extension in Windows


AnweshGangula:
I'm running the build command (`py build --chrome`) in my Windows PC and I'm getting the following error

```
Could not find platform independent libraries <prefix>
Traceback (most recent call last):
  File "C:\Users\AnweshGangula\Downloads\promnesia\extension\build", line 119, in <module>
    main()
  File "C:\Users\AnweshGangula\Downloads\promnesia\extension\build", line 64, in main
    check_call([
  File "C:\Users\AnweshGangula\AppData\Local\Programs\Python\Python311\Lib\subprocess.py", line 408, in check_call
    retcode = call(*popenargs, **kwargs)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\AnweshGangula\AppData\Local\Programs\Python\Python311\Lib\subprocess.py", line 389, in call
    with Popen(*popenargs, **kwargs) as p:
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\AnweshGangula\AppData\Local\Programs\Python\Python311\Lib\subprocess.py", line 1026, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "C:\Users\AnweshGangula\AppData\Local\Programs\Python\Python311\Lib\subprocess.py", line 1538, in _execute_child
    hp, ht, pid, tid = _winapi.CreateProcess(executable, args,
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
FileNotFoundError: [WinError 2] The system cannot find the file specified
```

AnweshGangula:
Turns out the python `./build` is not able to identify the executable. I tried to debug this, but could not spend too much time on this.

Instead I ran the `npm run build` script directly, but it didn't work since the parameter `target` in the webpack.config.js file was being passed by the python `./build` file. 

Debugging this was fairly simple

- I updated the webpack.config.js to read the argument from the terminal - `const target = env.TARGET ?? process.env.npm_config_target;`
- then I can run the command `npm run build --target=chrome` in the terminal to manually pass the target and for this to work

But this isn't a solution, rather a workaround.

AnweshGangula:
found the [solution](https://stackoverflow.com/a/28903670/6908282) for the issue with `./build --chrome`. the `check_call` command should specify `shell = True` for the command to work properly.

```
check_call([
    'npm', 'run', 'build',
[...]

PR #399, opened 2023-06-11
conditionally run extension build command in windows/linux

AnweshGangula:
as mentioned in Issue #398, this change conditionally runs the build command based on the user's OS

### DIFF
--- a/extension/build
+++ b/extension/build
@@ -3,6 +3,7 @@ import argparse
 import os
 from subprocess import check_call
 from pathlib import Path
+from sys import platform
 
 # right, specifying id in manifest doesn't seem to work
 # AMO responds with: Server response: Duplicate add-on ID found. (status: 400)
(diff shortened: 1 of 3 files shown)


extension/build
```diff
@@ -1,8 +1,8 @@
-#!/usr/bin/env python3
```
karlicoss: Hmm, I think that breaks the script (at least on linux/osx) because she bang is needed for the shell to run it as a python script. Can we keep it? I think you'll need to run it as `python3 extension/build` on Windows anyway, so it shouldn't hurt to keep it.

karlicoss (changes requested):
Sorry, somehow missed the notification! Thanks, it makes sense and looks good, just one comment because it breaks the github CI :)

extension/build
AnweshGangula: oh yeah that's okay.

karlicoss:
Hmm another CI error, but I think this is irrelevant to your changes, I'll investigate

karlicoss:
fixed the shebang issue since I had to rebase on top of latest master anyway -- everything looks good now. Thanks for the contribution!

PR #434, opened 2024-01-05
Use setup.py to install dependencies in docker

wilt00:
While building from the Dockerfile, I ran into some issues caused by pip resolving sqlalchemy to v1 instead of v2. This PR changes the Dockerfile to install directly from setup.py, rather than indirectly from a list of dependencies, to hopefully prevent issues like this from coming up in the future.

It looks like the reason we originally went with the individual dependencies approach is because setuptools-scm needs to get the version from the git history. Since we don't want to bloat the container by copying the entire git history into it, setuptools-scm can't find the version and the installation fails. Since then, it looks like setuptools-scm has added an override environment variable which we can use to pass in a version explicitly during the build.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -305,6 +305,7 @@ dmypy.json
 
 **/node_modules
 .idea
+.coverage*
 
 
 ######### NOTE: stuff above is from .gitignore
@@ -312,4 +313,12 @@ dmypy.json
 .git
 .ci  # otherwise build context is invalidated over and over again
 
+.circleci
+.github
+benchmarks
+ci
+doc
+docker/user_data
+docker/docker_files/indexer-config.py.example
+extension
 extension/dist
--- a/.dockerignore
+++ b/.dockerignore
@@ -305,7 +305,6 @@ dmypy.json
 
 **/node_modules
 .idea
-.coverage*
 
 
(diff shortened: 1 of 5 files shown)


.dockerignore
```diff
@@ -305,11 +305,20 @@ dmypy.json
...
 .ci  # otherwise build context is invalidated over and over again
 
+.circleci
+.github
+benchmarks
+ci
+doc
+docker/user_data
+docker/docker_files/indexer-config.py.example
+extension
```
karlicoss: I think best to keep these -- they are used (or can potentially be used by the CI). It fails now due to extension/.ci/build missing https://github.com/karlicoss/promnesia/actions/runs/7427827283/job/20327375203
https://github.com/karlicoss/promnesia/blob/master/.ci/end2end_tests.Dockerfile#L64

karlicoss:
Thanks! Sorry just noticed -- somehow didn't receive the email notification

.dockerignore
wilt00: I didn't think about the impact this would have on CI - I've now pushed changes to use a less lazy way of reducing the size of the docker image

wilt00:
Sorry for the drive-by PR @karlicoss ! I've finally fixed the CI issues here.

PR #451, opened 2024-06-19
fix(signal) update to recent db schema(11, Jun2024)

ankostis:
Reverse-engineered the changes in the db.

(Note, _Signal_ needs the latest `sqlcipher` executable.)


karlicoss:
thanks! sorry, just noticed the pr

PR #472, opened 2025-07-04
Update documentation to point correct location for obsidian and logseq sources

pjain02:
The documentation for both obsidian and logseq point to the `obsidian.py` and `logseq.py` files, whereas the correct path should be `auto_obsidian.py` and `auto_logseq.py`
Updated the documentation to reflect the same.


karlicoss:
thanks!

